Issue #2, opened 2017-07-27
Generic home page


srallen:
Move I2A specific content into its own components. Rework home page to be generic for selecting between I2A and Wildcam Darian Lab.

srallen:
Home page exists. Design and styling is a separate issue.

PR #9, opened 2017-07-27
Make home page generic

srallen:
For #2: 

This moves the intro2astro specific components into their own folder and makes the home page generic. I'm imagining a selection between the WCD:L and I2A areas. This sets a `projectCollections` array state that has the ids of the projects associated with these educational curricula. I'm open to a better idea of tracking that state. The home page will obviously need basic styling still. Perhaps some lorem ipsum placeholder to explain what the app is? Not sure if this home page really needs anything more. 

A follow up is to change the react router setup to use hash history instead of browser history so that direct links to either WCD:L or I2A areas will work. 

I also left a couple of todo comments inline.


srallen:
@shaunanoordin I'm gonna go ahead and merge this. If you have any comments, we can discuss in the slack channel or subsequent PRs.

Issue #5, opened 2017-07-27
Common header and footer


srallen:
Finish the grommet based common header and footer and move it to the ZRC repo.

PR #12, opened 2017-07-28
Finish footer

srallen:
For #5. Reworked the admin checkbox to be a controlled input. Finished the styles especially on smaller viewport sizes. Added the easter egg. ~~I need to write tests still.~~ Unit tests for the components are written. I'll write unit tests for the containers in a separate branch as this is getting to be a big PR.


src/components/layout/ZooFooter.jsx
```diff
@@ -35,48 +34,53 @@ const ZooFooter = (props) => {
...
       >
-        <Menu className="footer__nav" direction="row" inline={true}>
+        <Menu className="footer__nav" direction="row" inline={true} responsive={false}>
           {props.policyNavList.map((navItem, i) => createKeyedAnchorItem(navItem, i))}
         </Menu>
-        {props.adminContainer &&
-          props.adminContainer}
+        <div>
+          {props.adminContainer &&
+            props.adminContainer}
```
shaunanoordin: Query: why are there two `props.adminContainer`s here?

shaunanoordin:
## PR Review
* This PR adds the following:
  * Footer now has an Admin checkbox, available only when you're logged in as an admin. (Note: it appears I'm not an admin on staging, so I can't confirm this, but code looks fine.)
  * Further setup for mocha added; several components now have a `.spec.jsx` file for testing purposes. `npm run test` functions fine.

### Status
Code changes LGTM, although I can't test the Admin mode myself at the moment. Only thing before I merge is the question of the double `props.adminContainer` in `ZooFooter.jsx`

src/components/layout/ZooFooter.jsx
srallen: This is first checking for its existence and then renders it if it exists since that prop type is set to be a node. I could make this more explicit with a double bang for clarity:
```
{!!props.adminContainer &&
  props.adminContainer
```
Let me know if I should change it.

srallen:
@shaunanoordin what is you username on staging? I can make you an admin

shaunanoordin:
Thanks @srallen , my Zooniverse username is `darkeshard`

srallen:
@shaunanoordin I forgot that that checkbox is disabled for some reason in the UI. I did whitelist your user for uploads though. If you just ping someone with access to the staging Panoptes Rails console, they'll be able to make you an admin.

src/components/layout/ZooFooter.jsx
shaunanoordin: If the `props.adminContainer` is a node-type (or null or undefined), then I don't think the existential check is necessary. I tested what happened if we did, say...
```
render () {
  const ele = undefined;  //Also: null, false, 0, <span>Hello world</span>
  return {
    <div>The element is... {ele}</div>
  }
}
```
...and an `undefined`, `false`, or `null` value simply renders nothing, same if I did an existential check with `{ele && ele}`.

That said, if you believe the existential check is important, (maybe some React warnings I missed?) then I'm OK with merging this as is. 👍 

The `{!!props.adminContainer && props.adminContainer}` modification is unnecessary, unless for some bizarre reason the adminContainer somehow becomes the numeric value 0, which is the only condition that `{props.adminContainer && props.adminContainer}` doesn't catch properly.

shaunanoordin:
👍  Noted, thanks again @srallen !

src/components/layout/ZooFooter.jsx
srallen: I suppose I'm being overly careful about this since truthy/falsy behavior can be unintuitive in Javascript. You're right I can get away with just having `{props.adminContainer}` in the jsx. I'll update the PR.

PR #17, opened 2017-08-07
Example use of superagent with classrooms get request

srallen:
This is WIP PR of an initial setup of a duck for classrooms (using jumpstate conventions and helpers on top of redux) and using superagent with the superagent-jsonapify plugin. I moved the `config.js` file into a lib folder with another helper js file for the api calls. `edu-api.js` has some default headers and sets up the superagent-jsonapify plugin too. At some point, maybe not for the Sept 1 target date, but we could abstract out helper methods into this file to dry up some of the API calls. 

I didn't quite get this working as there are still some issues with staging. I've started on UI components on a separate branch, but thought I'd get an example of this up for you to look at first.

### DIFF
--- a/src/lib/edu-api.js
+++ b/src/lib/edu-api.js
@@ -1,16 +1,30 @@
 import superagent from 'superagent';
 import superagentJsonapify from 'superagent-jsonapify';
 import apiClient from 'panoptes-client/lib/api-client';
+import { config } from './config';
 
 superagentJsonapify(superagent);
 
 export const defaultHeaders = {
   'Content-Type': 'application/json',
-  'Accept': 'application/vnd.api+json',
-  'Authorization': apiClient.headers.Authorization
+  // 'Accept': 'application/vnd.api+json',
+  'Authorization': apiClient.headers.Authorization,
 };
 
-const eduAPI = superagent;
+export function get(endpoint) {
+  console.log('config.root', config.root)
+  return fetch(`${config.root}teachers/classrooms/`, {
+    method: 'GET',
+    mode: 'cors',
+    headers: new Headers({
+      'Authorization': apiClient.headers.Authorization,
+      'Content-Type': 'application/json'
+    })
+  }).then(response => response.json());
+  // return superagent.get(`${config.root}${endpoint}`)
+  //   .withCredentials()
+  //   .set(defaultHeaders)
+  //   .then(response => response);
+}
 
-export default eduAPI;
-window.eduAPI = eduAPI;
+window.eduAPI = superagent;
--- a/src/lib/edu-api.js
+++ b/src/lib/edu-api.js
@@ -5,26 +5,11 @@ import { config } from './config';
 
 superagentJsonapify(superagent);
 
-export const defaultHeaders = {
-  'Content-Type': 'application/json',
-  // 'Accept': 'application/vnd.api+json',
-  'Authorization': apiClient.headers.Authorization,
-};
-
 export function get(endpoint) {
-  console.log('config.root', config.root)
-  return fetch(`${config.root}teachers/classrooms/`, {
-    method: 'GET',
-    mode: 'cors',
-    headers: new Headers({
-      'Authorization': apiClient.headers.Authorization,
-      'Content-Type': 'application/json'
-    })
-  }).then(response => response.json());
-  // return superagent.get(`${config.root}${endpoint}`)
-  //   .withCredentials()
-  //   .set(defaultHeaders)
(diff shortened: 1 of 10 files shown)


srallen:
@shaunanoordin ok this is ready for review. As you know, staging is having a bit of an issue, but this works against production. Steps for testing:
- Be logged into panoptes as one of the testing accounts. I was using zootester1
- Set env using query param: http://localhost:3998/?env=production
- Browse to the Intro to Astronomy section
- There's no UI component in this branch, but you should see a successful network request for classrooms and the redux state set in dev tools.

src/lib/edu-api.js
```diff
@@ -0,0 +1,15 @@
...
+import apiClient from 'panoptes-client/lib/api-client';
+import { config } from './config';
+
+superagentJsonapify(superagent);
+
+export function get(endpoint) {
+  return superagent.get(`${config.root}${endpoint}`)
+    .set('Content-Type', 'application/json')
+    .set('Authorization', apiClient.headers.Authorization)
+    .then(response => response);
```
srallen: This should change to return `response.body.data`

src/lib/edu-api.js
```diff
@@ -0,0 +1,15 @@
+import superagent from 'superagent';
+import superagentJsonapify from 'superagent-jsonapify';
+import apiClient from 'panoptes-client/lib/api-client';
+import { config } from './config';
+
+superagentJsonapify(superagent);
+
+export function get(endpoint) {
```
srallen: We should add another argument to this helper method to take query params and build those into the request URL to later support things like filters.

shaunanoordin:
I've created an alternative implementation of the Education API Classrooms listing feature on PR #19 .

That PR has its own wall of text to discuss, so I only have one thing to comment on about this PR itself: the code LGTM, except for one thing...

```
get('teachers/classrooms/')
  .then((classrooms) => {
    Actions.classrooms.setClassrooms(classrooms);
    Actions.classrooms.fetchingClassrooms(false);
  }).catch((error) => {
...
```

...in that line of code, the first `then()` actually returns a HTTP response object, not a list of classrooms/the data object. We'd need an extra step here to check that `response.statusCode === 200` before proceeding with parsing the `response.body` for `setClassrooms()`.

Other than that, yeah! I can see the SuperAgent prodding the Education API (production) for data. 👍

srallen:
Agreed about checking the `statusCode`. We could check for `response.ok`  since I believe that indicates all successful responses (see: https://visionmedia.github.io/superagent/#response-properties). I also left an inline comment about returning `response.body.data` instead of the full response to the app. These can be done in the helper `get` method I wrote, so we don't have to repeat ourselves since we probably want to always do this.

srallen:
How do you want to merge our efforts? I can make the updates on handling the response and then you could rebase #19 from this branch?

src/lib/edu-api.js
shaunanoordin: I think returning `response` as is works better, actually. This would mean that whatever component/feature is calling the `get()` function will also be responsible for handling the error if, e.g. `response.statusCode === 404`.

If we want `get()` to return `response.body.data`, we'll need to bake error/bad response handling into this function, and it can get pretty unwieldy.

shaunanoordin:
@srallen I like how this PR is looking already, so this is my plan:
1. I'll merge this PR to master, as is (no need to change the response handling)
2. I'll update `classroom-api-v2` in PR #19 tomorrow morning to add the `classrooms.status` features and such.

Of note, when I merge PR #19:
* In PR #19, I'll need to change my response check from `response.statusCode === 200` to the simpler `response.ok`, since that's perfect for successful operations.
  * For future PRs, we can consider more sophisticated use of the `.statusCode` if we can figure out the specific errors that EduAPI returns, e.g. if it's a 401 we can prompt the Student to make sure she's registered for a class, a 404 then we prompt Teacher to check if they didn't delete the Assignment, etc etc.
* I'm in favour of keeping the `edu-api.js`'s `get()` method as is - returning `response` instead of `response.body.data` - since it means the component/feature/duck calling get() has to handle its own unique brand of errors.
* In PR #19, I'll also update `ClassroomManagerContainer` to reflect the updated `ducks/classrooms.js` , but please let me know if you'd rather I'd not touch that. (e.g. you've already got another PR improving the UI/visuals of the compo and don't want to worry about a merge-graine.)

srallen:
👌 regarding the error handling. I'll take a look at #19 and see if there would be any merge conflicts with what I'm doing. I'll try to get at least a WIP PR up for it today.

PR #19, opened 2017-08-08
Educaton API - Classrooms (ver 2)

shaunanoordin:
## PR Overview
* This is a basic/initial implementation of the Education API's Classrooms functionality.
  * This is based on, and is an alternate solution to, PR #17. Duplication of work, sorry!
* This PR has only one goal: connect to the EduAPI and get a list of Classrooms the logged in User (Teacher) owns.
screen shot 2017-08-08 at 20 19 39
_Here, in the /tester page, we can see (if you squint past the lack of CSS) the list of Classrooms for user zootester1 on production_
* There's also a tester page to make it easier to see what's happening, and to get a better idea of how the Redux states can play into the UI components.

Caveat for testing: at the moment, you need to test this on `production`, so you have to jump through a hoop or three.
1. Go to http://localhost:3998/?env=production.
2. Choose to Sign In
3. When you return from Panoptes, you should see that you're back on staging (http://localhost:3998/), so you need to hit Go to http://localhost:3998/?env=production once again. _You should now be logged in._

### Points of Discussion

This PR chooses different ways to implement the same functionality built in PR #17, and here are some points worth discussing.

**1. Using classrooms.status**

PR #17 uses the variables `classrooms.fetching=true/false` and `classrooms.error=object/null` to indicate the status of classrooms - whether data is being fetched and whether everything was done OK. So, e.g. `(classrooms.fetching === false)` would indicate the process is complete. (This is similar to current WildCam Lab.)

In my experience with WildCam Lab, I've found that determining the "status of a something" using variables like this can be problematic:
* It can be confusing to determine what's the current status: e.g. if `classrooms.classrooms===[] && !classrooms.fetching`, does that mean that you haven't started fetching data, or the fetch was successful but the data was empty `[]`?
* It's more work for devs to suss out why you're checking, e.g. `!classrooms.fetching && !classrooms.error`, to determine a success state.

I recommend explicitly spelling out the states using, e.g. `classrooms.status = CLASSROOM_STATUS.SUCCESS` as it's easier to map the flow of actions. The drawback though is that the code is bulkier, what with every component relying on `classroom.status` having to import `CLASSROOMS_STATUS`.

**2. Packaging CLASSROOMS_INITIAL_STATE and CLASSROOMS_PROPTYPES**

Using...
```
TesterHome.propTypes = {
  ...CLASSROOMS_PROPTYPES,
};
TesterHome.defaultProps = {
  ...CLASSROOMS_INITIAL_STATE,
};
```

Instead of the traditional...
```
TesterHome.propTypes = {
  classrooms: PropTypes.arrayOf(PropTypes.object),
  error: PropTypes.object,
  status: PropTypes.string,
};
TesterHome.defaultProps = {
  classrooms: [],
  error: null,
  status: CLASSROOMS_STATUS.IDLE,
};
```

This one's a newer pattern I'm experimenting with, so do let me know if you spot any Thermal Exhaust Port-levels of problems with the idea.

Basically, in my experience working with projects featuring multiple containers and components requiring shared props (usually connected by Redux), having to copy-paste the values for NewComponent.propTypes and NewComponent.defaultProps have been tedious and prone to unnecessary errors.

e.g. `store.responseCode` is changed from `string` to `number` with a default of `1` instead of `''`, and I need to update 5 different components. Then I forget about the 6th component and everything breaks.

Exporting out the initial states and propTypes should force every component to follow the _actual_ up to date data structure of the Redux store.

Foreseeable problems:
* The pattern doesn't do much for `mapStateToProps`
* What if, say, ComponentX needs to connect to the Classrooms store and the Assignments store, and both stores have the same variable names, like 'status' or 'students'? You'd need some spell-it-out-obviously code to differentiate the data to `classroomStudents` and `assignmentStudents`.

### Status
Ready for review. PR functionality does what it needs to do, but the implementation is subject to discussion.

### DIFF
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -15,6 +15,7 @@ export default function Home() {
     <Section colorIndex="light-2">
       <Anchor path="/astro" onClick={handleSettingProjectCollection.bind(null, config.astroProjects)}>Introduction to Astronomy</Anchor>
       <Anchor path="/" onClick={handleSettingProjectCollection.bind(null, [])}>Wildcam Darian: Labs</Anchor>
+      <Anchor path="/tester" onClick={handleSettingProjectCollection.bind(null, [])}>Tester</Anchor>
     </Section>
   );
 };
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -13,6 +13,7 @@ import AdminContainer from '../containers/layout/AdminContainer';
 import AdminLayoutIndicator from './layout/AdminLayoutIndicator';
 import AuthContainer from '../containers/layout/AuthContainer';
 import AstroHome from './astro/AstroHome';
+import TesterHome from './tester/TesterHome';
 
 const Main = ({ admin }) => {
   const mainHeaderNavList = [
@@ -34,6 +35,7 @@ const Main = ({ admin }) => {
           <Route exact path="/" component={Home} />
           <Route path="/about" component={AboutLayout} />
           <Route path="/astro" component={AstroHome} />
+          <Route path="/tester" component={TesterHome} />
         </Switch>
         <ZooFooter adminContainer={<AdminContainer />} />
       </Box>
--- a/src/components/tester/TesterHome.jsx
+++ b/src/components/tester/TesterHome.jsx
@@ -16,9 +16,6 @@ class TesterHome extends React.Component {
   }
 
   render() {
-    console.log('---'.repeat(80));
-    console.log(this.props.status === CLASSROOMS_STATUS.SUCCESS);
-    
     return (
       <div>
         <h1>CLASSROOMS</h1>
--- a/src/components/tester/TesterHome.jsx
+++ b/src/components/tester/TesterHome.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { Actions } from 'jumpstate';
+
+const TesterHome = (props) => {
+  return (
+    <div>
+      ...
+    </div>
+  );
+};
+
+TesterHome.propTypes = {};
+
+TesterHome.defaultProps = {};
+
+function mapStateToProps(state) {
+  return {};
+}
+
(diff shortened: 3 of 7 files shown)


shaunanoordin:
These are some excellent points, thanks @srallen ! My basic response to everything you said is a 👌 , with the following notes:

* re: mapStateToProps - Florp, yeah, let's not jump straight into the full Jumpsuit library yet, not so close to launch. I'm OK with working with mapStateToProps as we currently do.
* Sandboxing - I saw this sandboxing feature in action on Auth and Classrooms, and I like how it works. 👍 My apologies, my note on the _'foreseeable problems: can't differentiate between variable names'_ was unclear; I meant that in the context of the "exporting initial states and proptypes".
```
//For example, say both Classroom and Assignments have the 'status' variable.
TesterHome.propTypes = {
  ...CLASSROOMS_PROPTYPES,
  ...ASSIGNMENTS_PROPTYPES,
  //Then we need to make exceptions for the colliding names, and that's uncool, man. :/
  classroomStatus: CLASSROOMS_PROPTYPES.status,
  assignmentsStatus: ASSIGNMENTS_PROPTYPES.status,
}
```
I may be overthinking this, though! 🤷‍

shaunanoordin:
## PR Update
* This PR now needs to follow from the recently merged #17 
* Of note, on the TODO list:
  * This PR needs to be rebased
  * The check for response should just use `response.ok`
  * ClassroomManagerContainer needs to be hooked up to the updated `ducks/classrooms`

### Status
Update in progress! Adding the WIP tag

shaunanoordin:
## PR Update
* This PR has been rebased to match the recently merged PR #17 
* Also:
  * We're now using response.ok to check if a response is, well, OK.
  * ClassroomManagerContainer has been hooked up to the Classrooms duck.
    * The visual design is just a placeholder, of course; I was mostly interested in seeing how Grommet parts work in action.
    * As noted before, please let me know if you want me to revert this component to what it looks like after PR #17, in case there's any possible merge conflicts with another "improve component UI" branch that's being worked on. 

### Status
Ready for review!

PR #20, opened 2017-08-09
Classrooms view

srallen:
This is still very much WIP. I don't want to admit how long it took me to figure out the table (just FYI more than one `<tbody>` in a table is valid HTML!)

Here's what this PR does:
- Adds an assignments duck and conforms to the conventions decided on in #19
- Partially reworks the auth duck to conform to conventions. Still need to add prop types to it.
- Abstracts a common `<CurriculumHome />` component that can be reused for the WD section.
- I didn't abstract out a common signed in view since I wasn't sure if there would be need for differing views here.
- Added support for queries to the `get()` helper function in `edu-api.js`
- Added `console.error(error)` to the async Effects catch. Sometimes type errors are happening and get set into the state but don't print into the console. It's useful for these kind of errors to get printed to the console. As I noted in the review comment in #19 we need to continue to determine exactly what kind of errors get caught by the promise catch.

Further questions: 
- The API current doesn't support an include query param request for classrooms to include assignments. This could be an alternative to requesting assignments with a classroom id filter. Do we want this?
- I'm currently doing a minor transformation of the data from the assignments request to nest them in to a key/value pair of the classroom id: linked assignments. Do you think there's a better way to associate these two resources?

I'd like to spend some time tidying the css of the table to get it closer to the design, then I think it is ready for review: https://invis.io/ZWCCGVIQA#/240875082_01-Classroom_List

Further work to be done in separate PR:
- Tests for the API calls and continuing to improve error management
- How do we want to message errors in the UI? I think grommet's Notification component is a good candidate for this: https://grommet.github.io/docs/notification
- Possibly breaking apart the ClassroomManager further into a separate component/container for the assignments part.

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -5,7 +5,7 @@ import rootReducer from '../ducks/reducer';
 
 const createStoreWithMiddleware = applyMiddleware(
   CreateJumpstateMiddleware(),
-  createLogger(),
+  // createLogger(),
 )(createStore);
 
 export default function configureStore(initialState) {
--- a/src/styles/components/classroom-manager.styl
+++ b/src/styles/components/classroom-manager.styl
@@ -1,0 +1,13 @@
+.classroom-manager
+  &__table
+    background-color: white
+    border: solid thin GREY_1
+
+  .table
+    &__caption
+      bottom: -2.25em
+      font-weight: bold
+      position: relative
+      text-align: left
+
+    // &__headers
--- a/src/styles/components/classroom-manager.styl
+++ b/src/styles/components/classroom-manager.styl
@@ -1,7 +1,7 @@
 .classroom-manager
   &__table
     background-color: white
-    border: solid thin GREY_1
+    border: solid 2px #DEDEDE
 
   .table
     &__caption
@@ -10,4 +10,11 @@
       position: relative
       text-align: left
 
-    // &__headers
+    &__row-header
+      background-color: GREY_5
+      border-top: solid 2px #DEDEDE !important
+      border-bottom: solid 2px #DEDEDE !important
+
+    &__row-data:not(:last-child)
+      border-bottom: solid 2px #DEDEDE
+
--- a/package-lock.json
+++ b/package-lock.json
@@ -13256,15 +13256,15 @@
       "dev": true
     },
     "zoo-grommet": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/zoo-grommet/-/zoo-grommet-0.2.1.tgz",
-      "integrity": "sha512-b3a8nSGouuy8lNOkZfnHGZFr5+N+mjWVC8ANTMn7yvf6K3qKLJUJui6IlweOCRE/Ax/MZYsYXALxaVj1LTNkKA==",
+      "version": "0.2.2",
+      "resolved": "https://registry.npmjs.org/zoo-grommet/-/zoo-grommet-0.2.2.tgz",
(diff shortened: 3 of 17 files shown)


shaunanoordin:
Some thoughts:
* You're right, a common signed-in view is probably extremely unlikely. We already have enough Classroom/Assignment components that we can simply quickly Lego-build a Signed In page for specific projects with little worry.
* `get()` + queries = 👍  
* _"The API current doesn't support an include query param request for classrooms to include assignments ... Do we want this?"_ Although I know we can work without this feature, it does make a lot of sense to have it. I'd probably mark it down to the 'good to have' priority - I'd really like to have this, but I'd understand if other EduAPI features take priority for the end-month deadline. (e.g "select only Classrooms that are for Astro/Darien/etc")
* _"... currently doing a minor transformation of the data ... think there's a better way?"_  The only thing I can think of is updating the EduAPI itself with our new requirements, which has evolved sharply since the early nebulousness of Gorongosa. Another good to have; we'll need to write up a proper req list.
* `gromment/Notification` = 👍  I was looking at that compo too, and it looks like it was built just for what we need.
  * I was going to also add that we'd need a similar Notification compo for success messages as well as error messages, but then I remembered setting [grommet/Notifaction.status](https://grommet.github.io/docs/status) already covers this. Woot!

srallen:
@shaunanoordin I've tidied up some of the base table styles. I'm gonna remove the WIP label as this PR is getting big enough. Gonna branch off and work more on the UI and adding a create form for classrooms to keep this from getting any bigger.

src/store/index.js
```diff
@@ -5,7 +5,7 @@ import rootReducer from '../ducks/reducer';
 
 const createStoreWithMiddleware = applyMiddleware(
   CreateJumpstateMiddleware(),
-  createLogger(),
+  // createLogger(),
```
shaunanoordin: Reminder to future devs: re-enable this line of code to see `{prev state, action, next state}` Redux debug in the web console. It gets hella noisy but it's useful.

src/styles/components/classroom-manager.styl
```diff
@@ -0,0 +1,20 @@
...
+  .table
+    &__caption
+      bottom: -2.25em
+      font-weight: bold
+      position: relative
+      text-align: left
+
+    &__row-header
+      background-color: GREY_5
+      border-top: solid 2px #DEDEDE !important
```
shaunanoordin: Minor observation: I assume the `!important` here is to overcome Grommet's bizarrely specific `.grommetux-background-color-index--light .grommetux-table th, .grommetux-background-color-index--dark .grommetux-background-color-index--light .grommetux-table th { border-color: ... }`. (Three .classes strong in specificity, really?) While I'm mildly uncomfortable with `!important` in the CSS, I can understand the necessity when working with a UX framework.

Minor question: How come #DEDEDE doesn't get called GREY_4? 🤷‍♂️

srallen: Yep, that's what the `!important` is for. I don't like it either and there may be a better way to deal with it.

The naming is reflecting what its color index is in the zoo grommet theme. I added #DEDEDE to the grey color list and it's the fifth hex code. One of them is I think FOREGROUND? Confusing yes? I'd like to go through and do a bit of renaming at some point.

src/store/index.js
srallen: Oh I forgot I did this. It was getting a bit noisy for me.

shaunanoordin: No worries, as long as we know where the switch is, we can always re-enable it when we need it. 👍  Or add some fancy schmancy `if (process.env.ENABLE_DEBUGMODE === true) {...}` code. 😄

src/styles/components/classroom-manager.styl
shaunanoordin: Ahhah, confusing, yes, but no worries - we can always come back and do some renaming & general refactoring after launch. 👍  At this moment, important thing is that everything works nicely.

PR #22, opened 2017-08-14
Classroom create form, improved styles, and copy link functionality

srallen:
What this PR does:
- A few improvements to the classroom management table to get closer to the design
- Copy link functionality for the classroom join link
- Adds the Toast grommet component to show success of copying the join link. This component could be used to notify success of classroom creation and other actions. It auto-closes on its own. This is instead of the Notification component which has to be dismissed and is much larger in visual space. I'm open to using either/or/both, so let me know your thoughts. 
- A non-functional UI for the classroom create form. You can view it by clicking on the "Create Classroom" button

This can all be tested by using zootester1 using production database. 

Next steps are more style improvements and making the create form functional.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2237,6 +2237,14 @@
       "resolved": "https://registry.npmjs.org/cookiejar/-/cookiejar-2.0.6.tgz",
       "integrity": "sha1-Cr81atANHFohnYjURRgEbdAmrP4="
     },
+    "copy-to-clipboard": {
+      "version": "3.0.8",
+      "resolved": "https://registry.npmjs.org/copy-to-clipboard/-/copy-to-clipboard-3.0.8.tgz",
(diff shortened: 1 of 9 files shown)


srallen:
@shaunanoordin Thanks for the thorough review. For your questions:
- I think we need to confirm teacher registration this week for both I2A and WD as well as what the API expects. My guess is that it is not required for I2A. 
- Yes, there is a classroom details view that I haven't started. I imagine an edit button there that basically opens the same form that has inputs filled with current values. In that case, I can probably generalize the create form (I like to start with specific use cases then generalize if applicable)
- Agreed about the join link

Apologies for being unclear. I am working on the functionality on a separate branch, so this can merge. I'll likely have the functional form branch ready today.

shaunanoordin:
👍  on all of the above, and _(I like to start with specific use cases then generalize if applicable)_ gets an extra 👌  just for being sensible.

OK, let's merge this!

PR #23, opened 2017-08-15
WildCam Darien Map

shaunanoordin:
## PR Overview
* This PR adds the Map Explorer for the WildCam Darien portion of the Education API Front End.
* Work in progress:
screen shot 2017-08-15 at 18 41 54
_Currently viewable at the Tester page._

The Map Explorer requires the following features:
* A Teacher is able to view all the WildCam cameras on the maps of Darien/Soberania National Parks. 👍 
  * A Teacher is able to view data important to the WildCam project - e.g. camera markers should indicate number of species found at that location. The specifics are to be determined by what's made available with the data.
  * A Teacher is able to select 'Filters' to narrow down the type of data being shown on the map. 🔗
  * A Teacher is able to download the data being shown on the map, in CSV format. 🔗
  * A Teacher is able to select the data for the creation of WildCam Darien assignments. 🔗🔗🔗
* A Teacher is able to navigate through the map of Darien/Soberania National Parks, similar to how they can do so on Google Maps. 👍 

👍 indicates features that have been completed as of this post.
🔗 indicates features that may need to be split off into a separate PR.

### Status
Work in progress. Current focus is on cramming in as much functionality as possible, then cleaning up the code into proper components & etc, before removing that WIP label.

Please note I've added one temporary code change to expedite development: _HashRouter_ is used, to allow easy refreshing of the page. This should be reverted before merging. If I forget by the time I remove the WIP label, please holla.


srallen:
Do you want to just switch back to `HashRouter`? It was my mistake in forgetting the browser standards for the order of the query and the hash in the URL. It does make it easier for refreshing since we're developing against production right now and I've been doing it too! We can easily go back to `BrowserRouter` before we deploy... or not! I don't have a strong opinion either way.

src/components/tester/TesterHome.jsx
```diff
@@ -1,61 +1,27 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import MapExplorer from '../../containers/maps/MapExplorer';
 import { connect } from 'react-redux';
-import { Actions } from 'jumpstate';
-import {
-  CLASSROOMS_STATUS, CLASSROOMS_INITIAL_STATE, CLASSROOMS_PROPTYPES,
-} from '../../ducks/classrooms';
+
+const mapConfig = require('../../lib/wildcam-darien.mapConfig.json');
```
srallen: Webpack 3 has json-loader support built in. Did you have issues using new ES6 module import syntax for this? Just curious why you're using CommonJS require here.

shaunanoordin: _Picard face palm._ Thanks for catching that - I wasn't paying attention when I did some copy pasta from WildCam Gorongosa Lab - this should be changed to `import`

shaunanoordin:
@srallen I'm very much in favour of HashRouter for the ease of development, so I'm quite happy to keep on hashing. 👍  #hashtag

shaunanoordin:
re: `react-leaflet`, I remember trying out that library when building ol' WildCam Gorongosa Lab, but the problem was that I couldn't get React-Leaflet to integrate with the CartoDB library. However, it should be noted that was _many_ iterations ago, and I've moved waaay past tightly-coupling the MapExplorer with Carto's libraries since then.

I'm sticking with the 'vanilla' Leaflet implementation for now since it's faster for me, but `react-leaflet` is worth re-investigating if it makes more intuitive sense to other React devs, for future development & maintenance.

src/lib/wildcam-darien.mapConfig.json
```diff
@@ -0,0 +1,43 @@
+{
+  "database": {
+    "url": "http://wildcam-darien.carto.com/api/v2/sql?format=GeoJSON&q={SQLQUERY}",
+    "queries": {
+      "selectCameras": "SELECT * FROM cameras {WHERE}",
+      "selectCameraCount": "SELECT cam.*, COUNT(sbjagg.*) as count FROM cameras AS cam LEFT JOIN (SELECT sbj.camera, sbj.location, sbj.date, sbj.season, sbj.time_period, agg.data_choice, agg.subject_id FROM subjects AS sbj INNER JOIN aggregations AS agg ON sbj.subject_id = agg.subject_id) AS sbjagg ON cam.id = sbjagg.camera GROUP BY cam.cartodb_id ORDER BY count DESC"
```
shaunanoordin: I'm starting to wonder if this JSON file would work better as a .JS file, just so I can use multiline strings for clarity, and so I can actually add in some comments. Otherwise, this whole JSON config is a mystery wrapped in an enigma dressed in confusion and wearing a tiny hat that says "huh?"

src/containers/maps/MapVisuals.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+import L from 'leaflet';
+
+import {
+  MAPEXPLORER_INITIAL_STATE, MAPEXPLORER_PROPTYPES
+} from '../../ducks/mapexplorer';
+
+//Arbitrary values for a default map marker.
+//TODO / QUESTION: make this customisable?
+const DEFAULT_MARKER = {
```
shaunanoordin: Sigh, this is definitely something I need to revisit if I want to make the MapExplorer usable for non-Darien, non-Gorongosa projects.

shaunanoordin:
## PR Update
* We're now using **HashRouter** again, #hashtag
* MapExplorer now has better-defined subcomponents:
  * MapVisuals now only handles the visualisation. As it should. It doesn't care about project specifics.
screen shot 2017-08-17 at 20 33 21
_Unfortunately, some of those Soberania/Darien cameras are literally on top of one another._
    * TODO: clicking on a Map Marker should show users all photos from that Map Marker's camera. (Next next PR)
  * The generalised map data logic (i.e. connecting to a Carto or Carto-like database API) has been moved to the MapExplorer duck.
    * Reflection: I need to monitor that the MapExplorer duck is generic enough to support all survey/map-type data logic. Otherwise, I'd need to look at a project-specific WildCamDarienMap duck. 
  * All _project-specific (i.e. Darien-specific)_ data/logic/config now sits in `/src/lib/wildcam-darien.mapConfig.json`.
    * Reflection: I'm starting to think this should be a .JS file instead of .JSON...
  * MapControls is currently just a placeholder for future filter controls. (Next PR)
* Coding stuff: I've packaged all the MapExplorer parts into `/src/containers/maps`, but I'll need to review if it'll make sense to place MapControls and MapVisuals into `/src/components/maps` (where components like "Map Popup Modal" will eventually find a home).

One of the key points of this PR is to improve on the existing MapExplorer build from WildCam Gorongosa Lab. With this new MapExplorer, the feature needs to be versatile enough to support a variety of different projects, a variety of different data sources (e.g. Carto, or a Zooniverse-brand SQL server), and a variety of different map settings (e.g. different tile layers)

So far, this has been done by making sure the project-specific data/logic _only_ exists in a config file, but I'l have to see how far I can go with this arrangement.

### Status
@srallen this is now ready for review! To see the map in action, please go to http://localhost:3998/#/tester

shaunanoordin:
Next PR(s) will need to:
* Allow users to select filters (by species, etc) (MapControls.jsx)
* Allow users to download all the data based on said filters.
  * Let me know what you think of the [SuperDownloadButton](https://github.com/zooniverse/wildcam-gorongosa-education/blob/master/src/modules/common/components/SuperDownloadButton.jsx) as I'll be cribbing most of the logic from that compo (with mods to make it more generalised across all projects, of course). I'll welcome any suggestions, especially on 'proper' Grommet element structure if applicable.
* Haz algo en español

src/containers/maps/MapVisuals.jsx
```diff
@@ -0,0 +1,171 @@
...
+      </section>
+    );
+  }
+  
+  componentDidMount() {
+    this.initMapExplorer();
+  }
+  
+  componentWillReceiveProps(nextProps) {
+    this.updateDataLayer(nextProps);
```
srallen: Do you always want this to fire when new props are received by the component?

src/lib/wildcam-darien.mapConfig.json
srallen: Maybe it could be a class with a few helper class methods to build SQL queries and a constructor when initialized could set the url of the server?

src/styles/components/map-explorer.styl
```diff
@@ -0,0 +1,22 @@
+.map-explorer
```
srallen: Just to double up on the suggestion, the grommet `Box` is a div by default with flex box properties set. A class here still might be necessary for scoping any children classes for the component.

src/styles/components/map-explorer.styl
```diff
@@ -0,0 +1,22 @@
...
+  overflow: hidden
+  display: flex
+  position: relative
+  
+  .map-visuals
+    flex: 0 0 auto
+    width: 100%
+    height: 100%
+    
+    #mapVisuals
```
srallen: CSS classes or ids aren't typically camel cased. If this id is being used in javascript, it might be a good idea to decouple the styles and the javascript usage if possible. Use a CSS class instead and use a `data` property to target the DOM element? Or a React `ref`? I'm not very familiar with Leaflet's lib, though, so you can ignore this if using an id is necessary for its API.

src/containers/maps/MapExplorer.jsx
```diff
@@ -0,0 +1,46 @@
...
+class MapExplorer extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+  
+  //----------------------------------------------------------------
+  
+  render() {
+    return (
+      <div className="map-explorer" style={{height: '90vh'}}>
```
shaunanoordin: Sweet! This is going into the `wildcam-darien-map-v2` branch

src/containers/maps/MapVisuals.jsx
```diff
@@ -0,0 +1,171 @@
...
+    const cameraId = e.target.feature.properties.id;
+    alert(`TODO: Examine camera ${cameraId}`);  //TODO: Make this a "View a single camera's photos" action.
+  }
+  
+  //----------------------------------------------------------------
+  
+  render() {
+    return (
+      <section ref="mapVisuals" className="map-visuals">
+        <link
```
shaunanoordin: I used the 'component-specific <link>' method instead of adding the CSS link at index.tpl.html because I didn't want the CSS to load when people were only using Intro 2 Astro and not WildCam Darien... but now I realise I may be overthinking it somewhat, since the leaflet.css sits at a mere 14kb uncompressed.

OK, this gets moved to `main.css` with the `-v2` branch. Thanks for the `@import url()` trick; I wasn't aware this was a method of importing external CSS! Sweet!

src/containers/maps/MapVisuals.jsx
shaunanoordin: That's correct, it needs to be fired every time `this.props.markersData` is updated. This is because `updateDataLayer()` is the equivalent of a `render` method for the implementation of the Leaflet map.

The map doesn't natively listen to React prop changes natively, so this tells the map when to react. (This might be something `react-leaflet` solves, but I'd need to explore that implementation before I can comment confidently.)

shaunanoordin: OOOHHH I get now why it's called React!

src/styles/components/map-explorer.styl
shaunanoordin: Native Leaflet allows us to create a map either with an element's ID, or a reference to the HTML element itself. (http://leafletjs.com/reference-1.2.0.html#map-factory)

IIRC the reason I use IDs instead of an element reference is because the standard React ref method returns a React Element, which I'd then need to unpack into a standard HTML DOM Element, which I felt was messy. That, it's worth for me to re-check this assumption in the `-v2` branch.

Also: I should rename those `map-explorer.styl` classes anyway to keep those naming conventions consistent.

PR #24, opened 2017-08-16
Create form functionality and delete action

srallen:
This PR finishes the functionality for the create action as well as adds a delete button to the classroom list and the delete functionality.

- Improved UI for the create form modal
- A jumpstate Effect for getting classrooms and their linked assignments. This makes this sequence of API calls available to the whole app which is useful both for the delete method managed by the `ClassroomManagerContainer` and the create method managed by the `CreateClassroomFormContainer`.
- Related to the new Effect is a question about API usage optimization. We may want to consider alternatives to requesting for all classrooms again after create and delete. I left inline comments specifically about what I mean.
- Added `post` and `httpDelete` helper functions

### DIFF
--- a/src/containers/common/ClassroomCreateFormContainer.jsx
+++ b/src/containers/common/ClassroomCreateFormContainer.jsx
@@ -29,12 +29,18 @@ class ClassroomCreateFormContainer extends React.Component {
 
   onSubmit(event) {
     event.preventDefault();
+    // TODO: Add project id(s) associated to classroom create
     Actions.createClassroom(this.state.fields)
       .then(() => {
         Actions.classrooms.setCreateFormVisibility();
         if (this.props.projectCollection === config.astroProjects) {
           console.log('TODO: Auto create assignments for I2A');
-          // Actions.assignments.createAssignment()
+          // TODO: Actions.assignments.createAssignment().then(Actions.getClassroomsAndAssignments());
+          // For API optimization, we could merge the returned classroom into the local app state
+          // Then only call for the linked assignments for that one classroom
+          Actions.getClassroomsAndAssignments();
+        } else {
+          Actions.getClassroomsAndAssignments();
         }
       });
   }
--- a/src/containers/common/ClassroomCreateFormContainer.jsx
+++ b/src/containers/common/ClassroomCreateFormContainer.jsx
@@ -1,17 +1,67 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { Actions } from 'jumpstate';
 import ClassroomCreateForm from '../../components/common/ClassroomCreateForm';
+import { config } from '../../lib/config';
 
 class ClassroomCreateFormContainer extends React.Component {
   constructor(props) {
     super(props);
+
+    this.state = {
+      fields: {
+        name: '',
+        subject: '',
+        school: '',
+        description: ''
+      }
+    };
+
+    this.onSubmit = this.onSubmit.bind(this);
+    this.onChange = this.onChange.bind(this);
+  }
+
+  onChange(event) {
+    const fields = { ...this.state.fields, [event.target.id]: event.target.value };
+    this.setState({ fields });
+  }
+
+  onSubmit(event) {
+    event.preventDefault();
+    Actions.createClassroom(this.state.fields)
+      .then(() => {
+        Actions.classrooms.setCreateFormVisibility();
+        if (this.props.projectCollection === config.astroProjects) {
+          console.log('TODO: Auto create assignments for I2A');
+          // Actions.assignments.createAssignment()
+        }
+      });
   }
 
   render() {
     return (
-      <ClassroomCreateForm />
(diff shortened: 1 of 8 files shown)


src/containers/common/ClassroomCreateFormContainer.jsx
```diff
@@ -1,17 +1,73 @@
...
+    const fields = { ...this.state.fields, [event.target.id]: event.target.value };
+    this.setState({ fields });
+  }
+
+  onSubmit(event) {
+    event.preventDefault();
+    // TODO: Add project id(s) associated to classroom create
+    Actions.createClassroom(this.state.fields)
+      .then(() => {
+        Actions.classrooms.setCreateFormVisibility();
```
shaunanoordin: [minor, naming] I should've seen this in the last PR - setCreateFormVisibility() might make more sense as toggleCreateFormVisibility().

src/containers/common/ClassroomCreateFormContainer.jsx
```diff
@@ -1,17 +1,73 @@
...
+        if (this.props.projectCollection === config.astroProjects) {
```
shaunanoordin: Unfortunately I don't have the necessary project set up to test this `if ()` clause, but ah well, it's a TODO. I understand what's being tested here, I just need to check `Array() === []` equivalence because JS be cray cray.

shaunanoordin: Whoops, I forgot I could just check array equivalence in consoles; `[123] !== [123]` because technically the array variables are pointers to different objects.

Anyway, I imagine the final form of this will be something like `if (this.props.projectCollection.hasSomeMatchingElementsAsArray(config.astroProjects))`

src/containers/common/ClassroomManagerContainer.jsx
```diff
@@ -40,6 +36,15 @@ class ClassroomManagerContainer extends React.Component {
     this.setState({ toast: { message: null, status: null } });
   }
 
+  deleteClassroom(id) {
+    // TODO: Add alert to ask if user is sure
+    Actions.deleteClassroom(id).then(() => {
+      // TODO: For API optimization, do we want to instead manually remove the classroom
+      // out of local app state instead of making another API call
```
shaunanoordin: API optimisation: 👍 

That said, I wouldn't be _too_ worried about optimising the Create/Edit/Delete Classroom features since they wouldn't be called too frequently in a short period of time. Refreshing the list of Classrooms via API GET seems fine for now.

I'm trying to think of cases where update-local-state-instead-of-GET-from-API optimisation would be incredibly important, though... when adding Assignments? When Teachers start having so many Classrooms/Students/Assignments that individual GET calls become humongo?

src/ducks/classrooms.js
```diff
@@ -1,11 +1,13 @@
 import { State, Effect, Actions } from 'jumpstate';
 import PropTypes from 'prop-types';
-import { get } from '../lib/edu-api';
+import { get, post, httpDelete } from '../lib/edu-api';
```
shaunanoordin: [minor, irrelevant] I understand why this has to break the naming convention and be called `httpDelete` instead of `delete`, but it still makes me sad. 😑

src/containers/common/ClassroomCreateFormContainer.jsx
srallen: Yeah, I probably should have just commented this whole clause since it's non-functional at the moment. But the basic idea is that I'll want to do a comparison of the app state to the config and if the app state is set to the I2A projects, then auto create the relevant assignments on classroom create.

src/containers/common/ClassroomManagerContainer.jsx
srallen: We can leave the comments in for now to remind ourselves of where optimization can happen, but you're right in that it's not critical for the initial launch of the app.

src/ducks/classrooms.js
srallen: yeah, same. I tried to get away with it but the linter yelled at me.

src/ducks/classrooms.js
```diff
@@ -58,10 +67,54 @@ Effect('getClassrooms', () => {
...
+    });
+});
+
+Effect('getClassroomsAndAssignments', () => {
+  Actions.getClassrooms().then((classrooms) => {
+    classrooms.forEach((classroom) => {
+      // TODO: If many pages of assignments exist,
+      // loop through the number of pages to request all of the data
+      // and concatenate the response data together for the app state
+      // Neither Pagination nor infinite scroll would be good UX for current table design.
```
shaunanoordin: In practice, I've never seen a Teacher who so many Classrooms/Assignments that pagination becomes an issue. (Assumption: I2A wouldn't be too different in usage from WildCam Lab.) Personally, I'd put this pretty low on the ToDo/Optimisation list, but if we want to be sure, we can always ask Zach to check the average + max number of Classrooms/Assignments per Teacher so we can better anticipate the data load.

Issue #34, opened 2017-08-23
Manual Deploy Script could use work


shaunanoordin:
## Investigation Needed
This follows PR #33 and its extensive _list of known WTFs_

Issue 1:  What's wrong with ExtractTextPlugin in webpack.production.config.js?
* Original code:
```
test: /\.css$/,
use: ExtractTextPlugin.extract({
  fallback: 'style-loader',
  loader: 'css-loader',
  options: {
    includePaths: [
      path.resolve(__dirname, 'node_modules/zoo-grommet/dist'),
      path.resolve(__dirname, 'node_modules/zooniverse-react-components/lib')
    ]
  }
}),
```
* PR 33 removed `ExtractTextPlugin` to dodge an error. Is there a better way to solve this?

Issue 2: 💀  Why does the built & deployed website ( https://classroom.zooniverse.org ) thinks it's on a staging/development environment?

Issue 3: Why does the script for `npm run deploy-production` have to be so unnecessarily long and duplicate the script for `npm run build`?

Issue 4: Is there a better alternative to `publisssh`?

### Status
Open season for investigation. Only Issue 2 is considered a high priority item since it affects our end users.

Tagging @srallen into this; Sarah, I'd appreciate your insights into this when you're back.

srallen:
ExtractTextPlugin is used to extract the CSS into a css file rather than the default webpack behavior which is to inject it into a style tag in the head of the HTML. There's likely something else wrong, perhaps the configuration rather than the plugin itself. We'll want the plugin to work, so the main.css file builds correctly.

If I get a chance, I'll take a look at your other questions.

srallen:
Ok, ignore what I'm saying partly. The stylus configuration is handling the extraction, so it probably is correct to remove the plugin from the css loader configuration. I'll double check to confirm.

srallen:
Closed by #36

PR #36, opened 2017-08-24
Export env variables. Add plugin to babel production env.

srallen:
For #34  (maybe fix?)

I haven't actually tried the deployment, but the build looks good.
- ExtractTextPlugin isn't needed for the css-loader configuration because the css files are imported into the main.styl file which is then transpiled by the stylus configuration and that will handle extraction to main.css. I think this is something that needs fixing in zoo-reduxify too.
- I fixed the environment variable exports in the npm scripts. I added the export command so that the variable is put into your bash profile. Not sure if this will fix the production env when deployed since I didn't try that. But I noticed that now the webpack configuration file seems to be correctly running in production with the build task.
- I removed the redundant script in the deployment script and calling `npm run build` instead.
- Added the transform-object-rest-spread to babel production so that it knows how to transpile that for the production build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,8 @@
       "presets": [
         ["es2015", { "modules": false }],
         "react"
-      ]
+      ],
+      "plugins": ["transform-object-rest-spread"]
     },
     "test": {
       "presets": ["es2015", "react"],
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test": "NODE_ENV=development BABEL_ENV=test mocha mocha-setup.js $(find src -name *.spec.jsx) --compilers js:babel-core/register --require mock-local-storage || true",
     "eslint": "eslint .",
     "build": "export BABEL_ENV=production; check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
-    "deploy-production": "export NODE_ENV=production npm run build && publisssh dist zooniverse-static/classroom.zooniverse.org"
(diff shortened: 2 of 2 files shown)


shaunanoordin:
👍 Sweet, thanks @srallen ! I'll try deploying with this branch and post an update.

shaunanoordin:
## PR Review
* This PR fixes the major issues listed in #34 
  * Issue 1: Confirmed that ExtractTextPlugin isn't needed.
  * Issue 2: Deploy now makes sure that classroom.zooniverse.org runs on production, not staging. (Note: see Testing below)
  * Issue 3: The `deploy-production` script has been tidied up. (Note: see Testing below.)

### Testing

`npm run build` runs successfully, yaaasss.

`npm run deploy-production` _only_ runs successfully if `npm run build` was already run beforehand. This is because the script...
`"deploy-production": "export NODE_ENV=production npm run build && publisssh dist zooniverse-static/classroom.zooniverse.org"`
...is missing a `;` after the export statement, causing the `npm run build` to not be called properly. I don't know, it's all $bash sorcery to me.

Anyway, following the code of `build`, adding a `;` fixes the issue. I have successfully deployed the app to classroom.zooniverse.org and the website acknowledge the `production` environment (as tested when attempting to Sign In). Once again: yaaasss.

### Status
Pushing a quick `;` fix, then merging.

shaunanoordin:
## PR Review (Update)
* `;` 👍 

### Status
Merging.

PR #39, opened 2017-08-28
WildCam Darien Map part 5: Camera Viewer

shaunanoordin:
## PR Overview
* This PR adds the "Camera Viewer" functionality to the map.
* This means that a user can click on a Map Marker on the Map to view the camera trap photos from that specific camera.
screen shot 2017-08-28 at 02 33 34
* The Camera Viewer has a very basic pagination system. (Paging? Pagination?)

### Dev Notes
* Blurgh, we have a thumbnail generating system... but it only works if the photos are hosted on zooniverse.org, I think. I've built the thumbnailing functionality in, but it doesn't work with WildCam Darien's photos so... that Camera Viewer is blasting users with 6 hi-res images from the Darien National Park, scaled to tiny sizes. BLURGH, I say.
* I'm not sure if there's a Grommet/React ready pagination system we want to use - I saw the ToDo in the Classroom components but I'm unaware of a specific solution that we're aiming towards. Let me know if/when we do - I'll replace the current paging system the Camera Viewer uses, which is functional but fairly adhoc.
* The `getActiveCamera` Action, unfortunately, breaks the _"Map Explorer should be project agnostic; mapConfig should be project specific"_ pattern I've been holding on to for the past few updates. This is because the `WHERE camera = 'G07'` clause SHOULD be a project-specific config and not generalised across all projects; I'll need to revisit this once brain function more.
* A part of me tells me I'm being rather clumsy with the styling (see anything to do with colours or `.grommetux-...` overrides in `map-explorer.styl`) and am fighting unnecessarily with the Grommet framework, but another part of me thinks special overrides are necessary given the conflict between Grommet's generalised, spacious component designs and the Map Explorer's ridiculous explosion of data in a limited amount of screen real estate.

### Status
Ready for review!

NOTE: I may need to self-merge this before too long, as the next branch (WildCam Darien Map part 6: Translations) is contingent on all other parts being merged/finalised, and Part 6 itself is going to be a doozy. If I do prematurely merge this branch, please comment anyway - I'll put in minor fixes/suggestions as Part 6 and/or major fixes/suggestions to all the previous Map Explorer stuff in a PR after that.

### DIFF
--- a/src/components/maps/CameraViewer.jsx
+++ b/src/components/maps/CameraViewer.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Actions } from 'jumpstate';
+
+import Box from 'grommet/components/Box';
+import Button from 'grommet/components/Button';
+import Label from 'grommet/components/Label';
+import SpinningIcon from 'grommet/components/icons/Spinning';
+
+import {
+  MAPEXPLORER_INITIAL_STATE, MAPEXPLORER_PROPTYPES,
+  MAPEXPLORER_CAMERA_STATUS,
+} from '../../ducks/mapexplorer';
+
+class CameraViewer extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+  
+  render() {
+    return (
+      <Box className="camera-viewer">
+        <Box>
+        </Box>
+        <Box>
+        </Box>
+        <Button label="Close" onClick={()=>{ Actions.mapexplorer.resetActiveCamera() }}></Button>
+      </Box>
+    );
+  }
+}
+
+CameraViewer.defaultProps = {
+  ...MAPEXPLORER_INITIAL_STATE,
+};
+CameraViewer.propTypes = {
+  ...MAPEXPLORER_PROPTYPES,
+};
+
+export default CameraViewer;
--- a/src/components/maps/CameraViewer.jsx
+++ b/src/components/maps/CameraViewer.jsx
@@ -5,7 +5,10 @@ import { Actions } from 'jumpstate';
 import Box from 'grommet/components/Box';
 import Button from 'grommet/components/Button';
 import Label from 'grommet/components/Label';
+import List from 'grommet/components/List';
+import ListItem from 'grommet/components/ListItem';
 import SpinningIcon from 'grommet/components/icons/Spinning';
+import Notification from 'grommet/components/Notification';
 
 import {
   MAPEXPLORER_INITIAL_STATE, MAPEXPLORER_PROPTYPES,
@@ -15,25 +18,66 @@ import {
 class CameraViewer extends React.Component {
   constructor(props) {
     super(props);
+  
+    this.renderData = this.renderData.bind(this);
+    this.renderMetadata = this.renderMetadata.bind(this);
   }
   
   render() {
+    if (!this.props.mapConfig) return null;
+    
     return (
-      <Box className="camera-viewer">
-        <Box>
+      <Box className="camera-viewer" align="center">
+        <Box className="camera-metadata">
+          {this.renderMetadata()}
         </Box>
-        <Box>
+        <Box className="camera-data">
+          {this.renderData()}
         </Box>
-        <Button label="Close" onClick={()=>{ Actions.mapexplorer.resetActiveCamera() }}></Button>
+        <Button label="Done" onClick={()=>{ Actions.mapexplorer.resetActiveCamera() }}></Button>
       </Box>
     );
+  }
+
+  renderMetadata() {
+    if (this.props.activeCameraMetadataStatus === MAPEXPLORER_CAMERA_STATUS.FETCHING) {
+      return <SpinningIcon />;
+    } else if (this.props.activeCameraMetadataStatus === MAPEXPLORER_CAMERA_STATUS.ERROR) {
+      return <Notification message='ERROR' status='critical' />
+    }
+    if (this.props.activeCameraMetadataStatus === MAPEXPLORER_CAMERA_STATUS.SUCCESS &&
+        this.props.activeCameraMetadata) {
+
+      return (
+        <List>
+        {Object.keys(this.props.activeCameraMetadata).map(key => (
+          <ListItem
+            key={`camera-metadata-${key}`}
+            justify='between'
+            separator='horizontal'
+          >
+            <span>{key}</span>
+            <span>{this.props.activeCameraMetadata[key]}</span>
+          </ListItem>
+        ))}
+        </List>
+      );
+    }
+
+    return null;
+  }
+
+  renderData() {
+    return null;
   }
 }
 
 CameraViewer.defaultProps = {
+  mapConfig: null,
   ...MAPEXPLORER_INITIAL_STATE,
 };
 CameraViewer.propTypes = {
+  mapConfig: PropTypes.object,
   ...MAPEXPLORER_PROPTYPES,
 };
 
--- a/src/components/maps/CameraViewer.jsx
+++ b/src/components/maps/CameraViewer.jsx
@@ -16,9 +16,6 @@ import {
 } from '../../ducks/mapexplorer';
 
 const ITEMS_PER_PAGE = 6;
-const THUMBNAIL_SERVER_URL = 'https://thumbnails.zooniverse.org/';
-const THUMBNAIL_WIDTH = 320;
-const THUMBNAIL_HEIGHT = 200;
 
 class CameraViewer extends React.Component {
   constructor(props) {
--- a/src/components/maps/CameraViewer.jsx
+++ b/src/components/maps/CameraViewer.jsx
@@ -9,10 +9,12 @@ import Label from 'grommet/components/Label';
 import Layer from 'grommet/components/Layer';
 import List from 'grommet/components/List';
 import ListItem from 'grommet/components/ListItem';
+import Notification from 'grommet/components/Notification';
 import Tile from 'grommet/components/Tile';
 import Tiles from 'grommet/components/Tiles';
+import FormNextIcon from 'grommet/components/icons/base/FormNext';
+import FormPreviousIcon from 'grommet/components/icons/base/FormPrevious';
 import SpinningIcon from 'grommet/components/icons/Spinning';
-import Notification from 'grommet/components/Notification';
 
 import {
   MAPEXPLORER_INITIAL_STATE, MAPEXPLORER_PROPTYPES,
@@ -30,6 +32,8 @@ class CameraViewer extends React.Component {
   
     this.renderData = this.renderData.bind(this);
     this.renderMetadata = this.renderMetadata.bind(this);
+    this.renderDataPaging = this.renderDataPaging.bind(this);
+    this.changeDataPaging = this.changeDataPaging.bind(this);
   
     this.state = {
       page: 0,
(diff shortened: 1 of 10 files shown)


src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+      return <SpinningIcon />;
+    } else if (this.props.activeCameraDataStatus === MAPEXPLORER_CAMERA_STATUS.ERROR) {
+      return <Notification message='ERROR' status='critical' />
+    }
+    
+    if (this.props.activeCameraDataStatus === MAPEXPLORER_CAMERA_STATUS.SUCCESS &&
+        this.props.activeCameraData) {
+      const data = this.props.activeCameraData;
+      const items = [];
+      for (let i = this.state.page * ITEMS_PER_PAGE;
```
srallen: This loop totally works, but I might suggest considering a combination of `Array.slice()` to get the six data items you want and then `Array.map()` to iterate through for rendering:
```
const currentPageOfData = data.slice(startingIndex, ITEMS_PER_PAGE);
return (
  <Tiles>
    {currentPageOfData.map((currentItem, i) => {
      const fixedLocation = this.fixLocation(currentItem.location) // the regex test and replace
      <Tile key={`unique-key-${i}`}>
         <Anchor href={currentItem.location}>
            <img src={fixedLocation} />
         </Anchor>
      </Tile>
    });
  </Tiles>
);
```

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+            `${THUMBNAIL_SERVER_URL}${THUMBNAIL_WIDTH}x${THUMBNAIL_HEIGHT}/` +
+            location.replace(/^(https?|ftp):\/\//, '');
+        }
+        //----------------
+        
+        
+        items.push(
+          <Tile className="camera-data-item" key={`camera-data-${i}`}>
+            <Anchor href={data[i].location}>
+              <img src={location} />
```
srallen: This should have an alt property for accessibility. Maybe something like `alt="Image from X location"` where X is replaced with the location name or camera id? Whatever makes sense. If there are no descriptions that make sense, then an empty alt property, `alt=""`, can be used to let screen readers know to skip trying to describe the image. The default behavior without any alt specified is to just read out the file name, which almost always isn't helpful.

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+            {this.renderDataPaging()}
+            {this.renderData()}
+          </Box>
+          <Button label="Done" onClick={()=>{ Actions.mapexplorer.resetActiveCamera() }}></Button>
+        </Box>
+      </Layer>
+    );
+  }
+
+  renderMetadata() {
```
srallen: Maybe this can be abstracted out into its own functional child component?

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+          </ListItem>
+        ))}
+        </List>
+      );
+    }
+
+    return null;
+  }
+
+  renderData() {
```
srallen: Maybe this can be a functional child component?

src/containers/maps/MapExplorer.jsx
```diff
@@ -44,20 +50,37 @@ class MapExplorer extends React.Component {
...
+  ...MAPEXPLORER_PROPTYPES,
 };
 MapExplorer.defaultProps = {
   mapConfig: null,
+  ...MAPEXPLORER_INITIAL_STATE,
 };
-const mapStateToProps = (state) => ({});
+const mapStateToProps = (state) => ({
+  activeCameraId: state.mapexplorer.activeCameraId,
+  activeCameraMetadata: state.mapexplorer.activeCameraMetadata,
```
srallen: Since `CameraViewer` is a class, why not connect it to the redux store and directly set the props from the app state there? I think it's the only component using `activeCameraMetadata`, `activeCameraData` and the associated statuses, correct?

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+    return null;
+  }
+
+  renderDataPaging() {
+    if (this.props.activeCameraDataStatus === MAPEXPLORER_CAMERA_STATUS.SUCCESS &&
+        this.props.activeCameraData) {
+      return (
+        <Box className="camera-data-paging" direction="row" pad="none" alignSelf="stretch" justify="center" separator="horizontal">
+          <Button
```
srallen: Noticed this after I submitted the review, but the back button doesn't seem to work for me.

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+      const data = this.props.activeCameraData;
+      const items = [];
+      for (let i = this.state.page * ITEMS_PER_PAGE;
+           i < data.length && i < (this.state.page + 1) * ITEMS_PER_PAGE;
+           i++) {
+        let location = data[i].location || '';
+        
+        //Thumbnails!
+        //----------------
+        if (/zooniverse\.org/i.test(location)) {
```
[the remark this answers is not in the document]
srallen: (Or alternatively use the `Thumbnail` component available from ZRC)

shaunanoordin:
👍 Thanks for the quick review, Sarah! I'll get an update out in the next few hours.

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+    this.renderDataPaging = this.renderDataPaging.bind(this);
+    this.changeDataPaging = this.changeDataPaging.bind(this);
+  
+    this.state = {
+      page: 0,
+    };
+  }
+
+  componentDidMount() {
+    this.setState({
```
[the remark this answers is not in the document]
shaunanoordin: Good point, I was probably being improperly paranoid here about the `state` not resetting when a user clicks on another Map Marker/camera. Which, come to think of it, wouldn't have triggered properly anyway.

src/components/maps/CameraViewer.jsx
```diff
@@ -0,0 +1,188 @@
...
+      page: 0,
+    });
+  }
+  
+  render() {
+    if (!this.props.mapConfig) return null;
+    
+    return (
+      <Layer className="camera-viewer">
```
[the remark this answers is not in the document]
shaunanoordin: I originally removed the `closer` code because I couldn't get the Close button styled the way I wanted (I actually cribbed the `<Layer>` code from your CreateClassroomForm - thanks, btw!) but this might go back to my comments about 'fighting with the Grommet framework too much'. I'll try re-enabling the standard `closer` code and see if I can get the style working - it's better that the EAPIFE shares as consistent a UI as possible, after all.

src/components/maps/CameraViewer.jsx
shaunanoordin: 👌

src/components/maps/CameraViewer.jsx
shaunanoordin: 👌

src/components/maps/CameraViewer.jsx
shaunanoordin: Hot damn, I can't believe I overlooked this.

Actually, it's _theoretically_ possible that I can create a fairly descriptive alt text, for example: _"Photo from SOBERANIA NATIONAL PARK at DAYTIME containing approximately ONE JAGUAR, we guess"_

That would require me to review the SQL calls, the project config, and the way aggregations are recorded in the DB, so I'll need to mark that as a future feature. For now, I'll adjust it enough for a fairly simple description, which unfortunately, wouldn't be anywhere as useful for screen readers. ("A photo from camera G01" is accurate but doesn't provide any useful context.)

src/containers/maps/MapExplorer.jsx
shaunanoordin: Making CamerViewer connected to Redux makes sense, but my concept is that the main `MapExplorer` container should be responsible for the "overall/big picture" state of the Map feature.

That is to say, a new developer should be able to go into <MapExplorer /> and say, "Oh, right, that `if (activeCameraId)` clause means that `<CameraViewer />` is only available when an Active Camera is specified." Otherwise said dev would need to dig into CameraViewer to see that it renders null when activeCameraId is null.

Functionally, it's really the same, so if you feel strongly about making CameraViewer connected to Redux then I'll be happy to make the changes. My feelings for keeping it as is, is about 7/10.

src/components/maps/CameraViewer.jsx
shaunanoordin: NANDEYO? Thanks for the catch, I'll look into this.

To clarify, you mean the `<` previous arrow key doesn't work on Pages 2, 3, 4, etc, correct? I'll dive in.

src/components/maps/CameraViewer.jsx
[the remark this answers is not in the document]
shaunanoordin: I've switched to using the ZRC Thumbnail, which I believe is the way to go moving forward, but two caveats: 1. for WildCam Darien photos specifically, I'm seeing 415 errors (similar to what caused the 'Blurgh' comments from the initial post) so the Thumbnail is switching to default  tactics and 2. ZRC Thumbnails need to be updated to allow alt tags, I think. 

Not much can be done about 1, but 2 is a definite future ToDo

src/components/maps/CameraViewer.jsx
srallen: Yeah, button to go to the previous page. I wasn't seeing any errors thrown in console, so it's probably an issue with the math in the handler.

shaunanoordin:
## PR Update
* CameraViewer's Grommet Layer code/style has been standardised; no more 'Done' button to close the page, just a the regular Close button seen in other parts of the EAPIFE.
* CameraViewer component has its functionality broken into CameraViewerMetadata, CameraViewerData, CameraViewerPaging.
  * Paging code has been improved, although there's still room for more improvements, I think.
  * NOTE: The CameraViewerPaging is NOT the suggested '(Really) Generalised Paging for the EAPIFE' component that @srallen suggested; if I'm building a (really) generalised component that's going to be used in, e.g. the ClassroomManager, I want to approach it in its own PR and with some Design to back me up. I'd be happy to see the code reused though!
* The CameraViewerData now uses the ZRC Thumbnail component, but there are some caveats there.
  1. WildCam Darien photos will cause 415 errors during thumbnailing attempts, so the Thumbnail compo will fall back to the original high-res images anyhoos. Still, the ZRC Thumbnail component is the right way to go forward.
  2. By default all Thumbnail items will have `alt=""`; there's room for improvement here as each photo can have a pretty descriptive alt text. This should be marked for a future feature upgrade.

### Status
Ready for review!

src/containers/maps/MapExplorer.jsx
srallen: I think `activeCameraId` makes sense to remain in `MapExplorer` since it is handling the logic for the render. What I'm suggesting is the data that is specific to the `CameraViewer` directly be mapped fro the redux store, `activeCameraMetadata` and `activeCameraData`.

Just coming back from the React Rally conference and there's a lot we're not doing to improve performance in our apps at the moment (like correctly use `shouldComponentUpdate` for example). One suggestion was to make sure smart components only have access to the data they need.

 I don't feel too strongly about this at the moment, however, something to consider for future optimization.

shaunanoordin:
Thanks @srallen ! Give me a minute to rebase PR #40  to the new master.

PR #40, opened 2017-08-28
WildCam Darien Map part 6: Translations

shaunanoordin:
## PR Overview
* This is a super basic, brute force version of a translation system, _specific_ for the WildCam Darien Project.
* The requirements are:
  * The WildCam Darien portion of the Education API Front End (i.e. the Map Explorer) must have the ability to switch between English and Spanish.
  * It must be ready by launch day.
* And that's about it. A more solid, well-thought out solution _needs_ to be implemented once we have more than three days to think about it.
* Unfortunately Spanish is not one of the languages I invested skill points in, so _no tengo idea de lo que estoy diciendo._ I welcome everyone's suggestions on how to improve `zooniversal-translator.es.js`.

### Status
WIP

This will be rebased to the new master branch once PR #39 has been merged.


shaunanoordin:
## PR Update
* Translations! The Zooniversal Translator works with a BUCKETLOAD of project-specific hacks.
* I'm not terribly proud of the hacks, but it delivers what the users require within the timeframe.
* Long term, I need to look at how I can reconcile a rather complex project config (well, mapConfig) system with an equally complex (proper) translation system. i.e. how do we handle a _generalised solution_ for the EAPIFE having multiple projects with their own specific configurations, when each project can have their own set of languages, without making the interplay of systems too convoluted for the standard human headspace?

### Status
Ready for review, but please note that it'll make more sense if we first...
1. merge PR #39 , then
2. rebase this branch onto the new master,
...because otherwise you'd need to factor in commits `9ceccb1` to `9ef02ef` when analysing the new code additions.

shaunanoordin:
## PR Update
* This PR has just been rebased to the new master, now that PR #39 has been merged.

### Status
Review-ready, @srallen !

src/components/maps/CameraViewerMetadata.jsx
```diff
@@ -21,16 +23,22 @@ const CameraViewerMetadata = (props) => {
...
-          key={`camera-metadata-${key}`}
-          justify='between'
-          separator='horizontal'
-        >
-          <span>{key}</span>
-          <span>{props.activeCameraMetadata[key]}</span>
-        </ListItem>
-      ))}
+      {Object.keys(props.activeCameraMetadata).map(key => {
+        console.log(key, ZooTran(key));
```
shaunanoordin: Wait what's this guy doing here

shaunanoordin:
Awesome, thanks @srallen !

Translations: I didn't translate the map layers because I originally didn't think it was necessary, but having it pointed out to me does show how the experience just looks weird. I'll patch in the extra map layer translations in the next PR instead though, because part 7 actually has even more layers (and map legends!) to be translated.

Camera Viewer Metadata List: that's functioning as intended, although 'intended' could use a second design pass, if I'm being honest. The metadata list in the CV is no longer full width because both the list and the photo container are sized according to content; the containers have max widths but no min-widths.

Issue #49, opened 2017-09-08
Finish routes for programs


srallen:
Finish route clean up for the integration of the programs. Since the program, I2A and WD:L, will be represented as a resource on the API, we can fetch for them and use a param in the URL, `:program`, to know what the program scope is for the views.

srallen:
Once zooniverse/education-api#62 is done then we can add the requests for programs on the home page and request for a specific program if a user navigates directly to a program specific page.

PR #52, opened 2017-09-14
Programs

srallen:
For #49 

@zwolf refactored the new projects model on the Education API into programs since that better represented what we're actually trying to do and also allows I2A to have a single classroom for the three assignments that will be auto-created on classroom creation.

This PR is a start to adding in meaningful routes that use the program, Wildcam Darien Lab or Astro 101 with Galaxy Zoo (Gonna be hard not to keep calling it I2A!). Right now this PR is still using some mocks I created in the programs' duck, however, this will be refactored to request for programs from the API instead once those endpoints are added.

I'm going to hold off on merging this because I added a landing page for the route `\wildcam-darien-lab` that is similar to the current Wildcam Lab's home page asking if you are an educator or an explorer. The route `\wildcam-darien-lab\students` is now what loads the `MapExplorer` component. I'd like to get feedback from @shaunanoordin before merging. 

I had to add some container components that I'm not totally sure is necessary, but it's how I got the routes to work properly with React Router v4. React Router v4 doesn't do nested routes anymore, at least not how you or I would understand them.

Still to do, but will branch off this and continue work separately as this PR is fairly large:
- Add program filter for classrooms
- Add program id on create for classrooms
- Auto-create assignments task for I2A

Questions: Do we need to add a filter for classrooms by program on the original Wildcam Lab? I'm concerned that some users from that will also use Wildcam Darien Lab and the current Wildcam Lab will show them classrooms from both since it won't be filtering. Perhaps we can add a filter there to show only classrooms that aren't linked to a program and later back fill the association when Wildcam Lab migrates to this new front end?


src/containers/common/HomeContainer.jsx
```diff
@@ -1,13 +1,19 @@
...
+
 import Home from '../../components/common/Home';
 
 class HomeContainer extends React.Component {
   constructor(props) {
     super(props);
   }
 
+  componentDidMount() {
+    // Actions.getPrograms();
```
shaunanoordin: Is this a placeholder for future functionality, or a feature that was in test but no longer used?

src/containers/common/ProgramHomeContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+      Actions.getProgram({ programs: this.props.programs, param: this.props.match.params.program });
+    }
+  }
+
+  componentWillUnmount() {
+    Actions.programs.selectProgram(PROGRAMS_INITIAL_STATE.program);
+  }
+
+  render() {
+    // How do I pass props down consistently? They only get initial props, not the updated ones.
```
shaunanoordin: I accidentally asked this in PR #57 instead of this PR, but essentially: which updated props were you trying to pass down, but this component didn't register?

src/ducks/programs.js
```diff
@@ -35,10 +35,10 @@ const i2a = {
 };
 
 const darian = {
```
shaunanoordin: Dangit Darian, go home and bring back your brother Darién!

srallen: hahah you have no idea how many time I kept getting it wrong while working on this.

src/ducks/programs.js
```diff
@@ -142,6 +142,55 @@ Effect('getProgram', (data) => {
     }).catch(error => handleError(error));
 });
 
+// For a UI managed by Admins eventually
```
shaunanoordin: Since there's no UI for these Actions _(Effects? Darn this terminology)_ or test code, have we checked if they function properly? I'm OK with merging these if they're not tested, but I'd feel better if we mark them with a "Theoretically should work, check later when UI's built".

shaunanoordin:
> I'm going to hold off on merging this because I added a landing page for the route \wildcam-darien-lab that is similar to the current Wildcam Lab's home page asking if you are an educator or an explorer. The route \wildcam-darien-lab\students is now what loads the MapExplorer component.

I'm super OK with this 👍  WildCam Darien Lab is supposed to mimic WildCam Gorongosa Lab, so not only does this make sense, and it's actually the intended way Darien Lab should be set up for Phase 2.

> Do we need to add a filter for classrooms by program on the original Wildcam Lab?

We absolutely do. There are two options here: either we (as suggested) make some light modifications to lab.wildcamgorongosa.org so that, after fetching a list of Classrooms, it only filters for Classrooms with no attached Projects;

OR, we make some significant changes to lab.wildcamgorongosa.org to better support Projects _and_ we retroactively attach a Gorongosa Project to all Classrooms with no projects (which we can infer to be Gorongosa Classrooms).

I suspect that the _eventual intent_ is that WildCam Gorongosa Lab - being the prototype for education stuff - _should_ be brought into Zooniverse Classrooms. On one hand, this is unlikely to happen in practice (due to time & resources) so a big update to WCGorongosaLab may not make sense. On the other hand, there are still benefits to making WC Gorongosa Lab future-compatible for such an eventuality.

For _practical_ purposes, I vote for making the minimum required changes to lab.wildcamgorongosa.org

src/ducks/programs.js
srallen: Yeah, they are untested and theoretically should work, but we don't need to test that until it is needed. I'll add a comment in a future PR.

src/containers/common/ProgramHomeContainer.jsx
srallen: I was trying to pass down props using the `Route` component:

`<Route {...props} path="/astro-101-with-galaxy-zoo/educators" component={AstroHome} />`

But what happened was that the child component at that route only got the default props, not the updated props once the selectedProgram was updated in the app store. I was hoping to be able to pass the prop to the child route like this so that I wouldn't have to connect both the container component and the child component to the store to get basically the same data. 

Either I don't understand how React Router v4 Route components work and this isn't supposed to work or it's a bug, but I didn't invest the time to figure that out and instead just have both the parent and child connected to the redux store. I left the comment to eventually come back to that and research in the docs and github issues to understand if this should work.

src/containers/common/HomeContainer.jsx
srallen: Yes! Programs still need the show route on the API so GET requests work: https://github.com/zooniverse/education-api/issues/62

I'd like to be able to request for the programs for the home page listing and then for the selectedProgram if a user navigates directly to I2A or Wildcam Darien. The mocks in the programs duck shows what I'm envisioning as the data we get back from the EduAPI.

srallen:
Ok, sounds good for the existing Wildcam Lab. I'll open an issue there about adding a filter for classrooms that are not linked to a program.

PR #56, opened 2017-09-15
Auto create assignments

### LINKED ISSUE #55: Auto-create the three standard assignments on classroom create for I2A

srallen:
This is a WIP branch to eventually close #55. This adds the `createAssignment` Effect in the assignment duck as well as some code to auto-create assignments if the program (like I2A) needs to do that. Right now I have some mocks for programs, but we should create these on the staging server. I'm planning on storing a lot of information in the metadata object on the program to keep track of the data needed for the assignments that will be auto-created.

It also should be noted, I branched this off of the `programs` branch, so this will need a rebase as soon as #52 is merged.

I'm not sure if this works yet as there's been issues with the education-api-staging server again this past week that hopefully will be addressed next week (noted in the slack channel; likely issue on docker swarm?). I'm depending on the program resource to exist and we don't want to deploy that to production yet, so we have to use staging here for development.

### DIFF
--- a/src/containers/classrooms/ClassroomsTableContainer.jsx
+++ b/src/containers/classrooms/ClassroomsTableContainer.jsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { Actions } from 'jumpstate';
+import AstroClassroomsTable from '../../components/astro/AstroClassroomsTable';
+
+import {
+  CLASSROOMS_INITIAL_STATE, CLASSROOMS_PROPTYPES
+} from '../../ducks/classrooms';
+import {
+  ASSIGNMENTS_INITIAL_STATE, ASSIGNMENTS_PROPTYPES
+} from '../../ducks/assignments';
+
+class ClassroomsTableContainer extends React.Component {
+  selectClassroom(classroom) {
+    Actions.classrooms.selectClassroom(classroom);
+  }
+
+  render() {
+    if (this.props.selectedProgram && this.props.selectedProgram.slug === '/astro-101-with-galaxy-zoo') {
+      return (
+        <AstroClassroomsTable
+          assignments={this.props.assignments}
+          assignmentsStatus={this.props.assignmentsStatus}
+          classrooms={this.props.classrooms}
+          deleteClassroom={this.props.deleteClassroom}
+          match={this.props.match}
+          maybeDeleteClassroom={this.props.maybeDeleteClassroom}
+          selectClassroom={this.selectClassroom}
+          selectedProgram={this.props.selectedProgram}
+        />
+      );
+    }
+
+    // TODO return the darien style classroom table
+    return null;
+  }
+}
+
+ClassroomsTableContainer.defaultProps = {
+  ...ASSIGNMENTS_INITIAL_STATE,
+  ...CLASSROOMS_INITIAL_STATE,
+  selectClassroom: () => {}
+};
+
+ClassroomsTableContainer.propTypes = {
+  ...ASSIGNMENTS_PROPTYPES,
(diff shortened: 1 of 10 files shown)


srallen:
Ok, this should be ready for review now. The staging server is fixed and so the auto-create for I2A classrooms work. I also created a new `ClassroomsTableContainer` and moved the table code in a project specific `AstroClassroomsTable` component. `ClassroomsTableContainer` will handle which table to load because I2A classrooms will not have any UI for assignment creation or editing. Darien classrooms won't have exports from caesar. These differences made it reasonable to separate out the table component.

I'm imagining similar table for Darien, but there could a button in the placeholder row for classrooms that don't have assignments yet to open up a modal form for creation then the assignments that do exist could have the pencil edit button next to their name and a X button at the end to delete similar to classrooms.

I'm still a little unsure on the organization of `ClassroomManagerContainer` vs `ClassroomsTableContainer` and I think it could use some more thought, but maybe for later PR.

Anyway to test this, you should be able to go to the classroom manager for I2A classrooms and create a new classroom which should trigger the auto-creation of the three standard assignments for their curriculum. 

Just to re-iterate, this should only merge after merging #52 and a rebase.

srallen:
This is now rebased and I fixed the spelling of darien in the program duck.

src/containers/classrooms/ClassroomsTableContainer.jsx
```diff
@@ -0,0 +1,61 @@
...
+  ASSIGNMENTS_INITIAL_STATE, ASSIGNMENTS_PROPTYPES
+} from '../../ducks/assignments';
+
+class ClassroomsTableContainer extends React.Component {
+  selectClassroom(classroom) {
+    Actions.classrooms.selectClassroom(classroom);
+  }
+
+  render() {
+    if (this.props.selectedProgram && this.props.selectedProgram.slug === '/astro-101-with-galaxy-zoo') {
```
shaunanoordin: This `.slug` makes me wonder if there's a better way we can organise our Projects; perhaps we need to consider organising Program-specific variables into a Program.config file, then place all those Program-specific config and components files into their own module-folder?

_(This is a minor comment for future discussion, and does not affect the review of this PR.)_

shaunanoordin: e.g. at the moment the `.slug` value (`/astro-101-with-galaxy-zoo`) exists as a (temporary?) const/config in `src/ducks/programs.js`, so that's a prime candidate for putting that into a `/lib/i2a/project.config` file so we can ensure consistent values across all components used by the Program.

shaunanoordin: EDIT: I just properly digested the new updates to `src/ducks/programs.js` - the Duck IS the Program config file now! Sorry, missed that.

In any case, that means I need to change my comment to: we really should start using the config values we've set up instead of hard-coding string values. This remains a minor comment, though.

shaunanoordin:
> I'm imagining similar table for Darien, but there could a button in the placeholder row for classrooms that don't have assignments yet to open up a modal form for creation then the assignments that do exist could have the pencil edit button next to their name and a X button at the end to delete similar to classrooms.

👌  The idea sounds legit. It should be fairly straightforward to create a Create Assignment modal form.

Although, that said, I need to double-check on the following _intended_ rule: that _all students of the classroom_ are automatically associated with (i.e. assigned to) _every assignment._ IIRC, WildCam Gorongosa Lab required manual assignments of Students to each Assignment in the Classroom, which in practice didn't make too much sense.

> I'm still a little unsure on the organization of ClassroomManagerContainer vs ClassroomsTableContainer and I think it could use some more thought, but maybe for later PR.

I imagine I'll have more to comment about this when I start creating src/components/DarienClassroomsTable

PR #60, opened 2017-10-04
Add I2A content. Style clean up. Fix routing bug

srallen:
This PR does a few things:
- Updates `grommet` to 1.8.2 and adds `classnames`
- I replaced some placeholder text with the actual text that will be used for I2A
- Reworked the classes used by `ProgramHome` and added styles specific to `AstroHome`
- Added the URL for the I2A Instructor's Guide
- Fixed a routing bug with the header links between the I2A and Wildcam Darien Lab using

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5344,7 +5344,8 @@
     "function-bind": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/function-bind/-/function-bind-1.1.0.tgz",
-      "integrity": "sha1-FhdnFMgBeY5Ojyz391KUZ7tKV3E="
+      "integrity": "sha1-FhdnFMgBeY5Ojyz391KUZ7tKV3E=",
+      "dev": true
     },
     "function.prototype.name": {
       "version": "1.0.0",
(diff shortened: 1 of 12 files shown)


shaunanoordin:
> Fixed a routing bug with the header links between the I2A and Wildcam Darien Lab using

I'm not too sure what to make of this; was there a routing bug that prevented us from switching between I2A and Darien, and that was fixed to the code update in src/containers/common/ProgramHomeContainer.jsx ?

srallen:
@shaunanoordin I'm not sure how the routing bug appeared either, but it must have been a regression we missed at some point. The `componentWillReceiveProps` that I added in `ProgramHomeContainer` fixes it. Basically if you navigated to Wildcam Darien Lab then clicked to go to Astro 101 or vice versa, then the `selectedProgram` in the app store wasn't properly updating.

The odd thing is that I remember this working correctly before, so I'm not sure when/how this broke.

shaunanoordin:
Thanks for the explanation, Sarah! 👍  OK, that's cleared, so this PR is now goin' a-mergin'.

PR #61, opened 2017-10-05
Fix deployment of images

srallen:
The webpack production configuration was missing the plugin to copy static asset files. I've added `CopyWebpackPlugin` as a dependency and the configuration to the plugin option in the production configuration.

This can be tested by running the build on its own, `npm run build` and checking the `dist` folder for the images.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2269,6 +2269,92 @@
         "toggle-selection": "1.0.6"
       }
     },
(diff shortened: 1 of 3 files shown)


srallen:
@shaunanoordin I'm going to go ahead and merge this since I've already deployed the fix for the I2A instructors to be able to see the site with working images.

PR #63, opened 2017-10-05
Add caesar request

srallen:
This PR lays out the scaffolding for the Caesar export requests:
- A Caesar export duck with an Effect that requests from Caesar. This part isn't working yet since I'm waiting on the user group id to be serialized with classrooms in zooniverse/education-api#67
- A modal to show the status of the export as well as options on how to download it. ~~I still need to add the buttons for this.~~ Placeholder button for CSV download is now added.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12195,6 +12195,11 @@
         "readable-stream": "2.3.3"
       }
     },
+    "superagent-graphql": {
+      "version": "0.1.2",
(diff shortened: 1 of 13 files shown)


srallen:
@shaunanoordin I think this is ready for review. Most of this is UI placeholder and is non-functional until the updates to the API to have the Panoptes user group id.

PR #64, opened 2017-10-06
Fix regression with ConfirmationDialog on classroom delete

srallen:
I think this was a regression that came in with #56. In the reorganization I did, I think I accidentally removed the `ConfirmationDialog` used with the classroom delete button.

I did some minor improvements with the success handling of the delete requests for both classroom and student. We only want to show the success message if there's a response from the API.


srallen:
I'm going to go ahead and merge this because being able to successfully delete a classroom will be helpful for me to be able to implement some of the functionality that I was waiting on with the latest EduAPI PR for programs (zooniverse/education-api#67).

PR #65, opened 2017-10-10
Use zooniverse group id

### LINKED ISSUE #7: API requests for caesar
Need JS client for API get requests to Caesar for the aggregated or calculated data from the raw classifications for I2A for a given classroom.

srallen:
This is a follow up to #59 and #63 and closes #7. Now that the actual zooniverse group id is serialized with the classroom, I can correctly build the URL for the project and for the caesar export request. The caesar export request will now correctly and as expected 404 since no exports will exist yet.

### DIFF
--- a/src/components/astro/AstroClassroomsTable.jsx
+++ b/src/components/astro/AstroClassroomsTable.jsx
@@ -90,7 +90,7 @@ const AstroClassroomsTable = (props) => {
                 props.assignmentsStatus === ASSIGNMENTS_STATUS.SUCCESS) &&
                 props.assignments[classroom.id].map((assignment) => {
                   const projectUrl = (assignmentsMetadata && assignmentsMetadata[assignment.workflowId]) ?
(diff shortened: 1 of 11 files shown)


src/components/astro/AstroClassroomsTable.jsx
```diff
@@ -90,7 +90,7 @@ const AstroClassroomsTable = (props) => {
                 props.assignmentsStatus === ASSIGNMENTS_STATUS.SUCCESS) &&
                 props.assignments[classroom.id].map((assignment) => {
                   const projectUrl = (assignmentsMetadata && assignmentsMetadata[assignment.workflowId]) ?
-                    `${config.zooniverse}/projects/${assignmentsMetadata[assignment.workflowId].slug}/classify?group=${classroom.zooniverse_user_group_id}` :
+                    `${config.zooniverse}/projects/${assignmentsMetadata[assignment.workflowId].slug}/classify?group=${classroom.attributes.zooniverse_group_id}` :
```
shaunanoordin: Do we need an existential check for `classroom.attributes` or can we be guaranteed there'll at least be some sort of object there?

PR #66, opened 2017-10-10
Request programs

### LINKED ISSUE #49: Finish routes for programs

srallen:
Closes #49. We now have I2A and WD: Lab programs on the staging database. We can request them using the API. Note: the API show route for the programs is still authenticated, so you have to be logged in for this to work. I've opened zooniverse/education-api#69 about it.

Also, another idea on clearly communicating status: I've created the `GenericStatusPage`. This is a fall back page to show in case the server is down or the load time is really slow. I'm thinking some teachers may bookmark the I2A or Wildcam page and directly navigate to those pages. In case the status is still 'fetching' or 'error', The GenericStatusPage will display.

However, we have a lot of ways of communicating status. From Grommet's Toast and Notification components to now this. I may be going overboard with it.

### DIFF
--- a/src/components/astro/AstroClassroomsTable.jsx
+++ b/src/components/astro/AstroClassroomsTable.jsx
@@ -40,7 +40,7 @@ const AstroClassroomsTable = (props) => {
         {props.classrooms.map((classroom) => {
           // TODO update URL once we have staging/production hosts
 
-          const joinURL = `${window.location.host}/#${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
+          const joinURL = `${window.location.host}/#/${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
           // Can we get linked assignments with classrooms in single get request?
           // No, if we want this, then we need to open an issue with the API
           // TODO replace classifications_target with calculated percentage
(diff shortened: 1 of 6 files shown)


shaunanoordin:
The PR itself looks fine (noted on the you-gotta-be-logged-in-to-see-the-home-page clause), but now that I've had a chance to examine what the `/programs` endpoint provides, I have some observations/questions about the implementation.

1. A Program's description and other 'settings' are stored in the Edu API, which is good- it should be nice and editable this way.
2. However, some resources - e.g. background image - is defined in the EduAPI ( `data[1].attributes.metadata.cardImage: "astro-background.jpg"` ) but the resources themselves are stored [locally](https://github.com/zooniverse/edu-api-front-end/blob/master/src/images/astro-background.jpg) and that seems slightly jarring.
3. Certain components like AppHeader don't fully implement program details, but I can take care of that in a separate PR.
4. I'm just wondering which Program details/information/resources we're planning to 'outsource' to EduAPI (or external hosting service, in the case of images), and which we're keeping 'in-house'. (e.g. our Routes are still hardcoded with the Program slugs, which is inevitable since we don't have an 'Education Prorgam Builder' or equivalent to fully generalise Programs.)

I'm assuming most of these are works in progress, so I have no real issues to flag at the moment. I'll finish the full PR review soon.

Issue #26, opened 2017-08-16
Add Google Sheets API integration


srallen:

srallen:
As a follow up to #69, the Caesar CSV export is the file that should be sent, not a test blob.

PR #69, opened 2017-10-12
Google API integration

srallen:
This is for #26 as an initial pass at adding the Google API javascript client, gapi. Google doesn't provide their javascript client as an NPM package, so I've had to do some wonky workarounds that aren't exactly the React-way-of-doing-things, but hey it's working so far.

The new `GoogleSheetsExportButton` will add the gapi script tag, instantiate the client and the auth client, and also load the Drive API. 

When you click the button, a modal window should pop up asking to authenticate Drive API permissions for your Google account. Then, for testing purposes, I've created a blob to send over save as a spreadsheet file on your drive. This is still WIP because a file is created on drive, but it's nameless and 0kb, so either I'm not setting up the blob correctly, not sending it over correctly, or both. In theory what should be sent over is the caesar CSV export not a on the fly created blob.

I did some slight renaming for the existing export modal and style tweaking to the SuperDownloadButton. 

I also need to add success (show toast sending file is successfully exported) and error handling.

One question with the authentication with Google is that on component unmount, I sign out of Google. This seems like it'd be good for security purposes, but it might get annoying to the user if they're exporting data often. We should see how the behavior of the authentication goes after you authenticate the first time. On the Google side of things, it should keep track which apps you've given permission to and unless you revoke access in your account settings, it should stay authenticated that way.

### DIFF
--- a/src/components/astro/GoogleDriveExportButton.jsx
+++ b/src/components/astro/GoogleDriveExportButton.jsx
@@ -14,7 +14,7 @@ const STATUS = {
 
 let GoogleAuth = null;
 
-class GoogleSheetsExportButton extends React.Component {
+class GoogleDriveExportButton extends React.Component {
   constructor() {
     super();
 
@@ -33,6 +33,12 @@ class GoogleSheetsExportButton extends React.Component {
     this.setupGoogleClient();
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextState.isAuthorized !== this.state.isAuthorized) return false;
+
+    return true;
+  }
+
   componentWillUnmount() {
     if (this.state.isAuthorized) GoogleAuth.signOut();
     const script = document.querySelector('#google-api');
@@ -49,7 +55,6 @@ class GoogleSheetsExportButton extends React.Component {
   }
 
   setupGoogleClient() {
-    console.log('gapi', window.gapi)
     if (!document.querySelector('#google-api')) {
       this.setState({ status: STATUS.CONFIGURING });
       const script = document.createElement('script');
@@ -66,7 +71,6 @@ class GoogleSheetsExportButton extends React.Component {
             discoveryDocs: config.googleDiscoveryDocs
           }).then(() => {
             GoogleAuth = gapi.auth2.getAuthInstance();
-            window.googleAuth = GoogleAuth;
             GoogleAuth.isSignedIn.listen(this.updateSigninStatus)
 
             gapi.client.load('drive', 'v3', () => {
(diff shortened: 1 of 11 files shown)


srallen:
Just an update on the auth behavior: Google is correctly only prompting for permission to your drive once, however, I do see the pop up each time asking which account I want to use. We probably do want that to prompt each time.

srallen:
Re: only an empty file being created. I'm not hitting the right end point for actual file creation. The javascript client's `gapi.client.drive.files.create()` doesn't do what you think it does 🤦‍♀️

srallen:
Update: I've added success and error handling. I need to add UI for the exporting as it's happening still. 

The POST is now working. A Google spreadsheet file is created now on the drive of the Google account you select to auth with (Thanks to a helpful stackoverflow example). Now I just need to figure out how to get the actual CSV data in the new file correctly.

src/components/astro/GoogleDriveExportButton.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  setupGoogleClient() {
+    if (!document.querySelector('#google-api')) {
+      this.setState({ status: STATUS.CONFIGURING });
+      const script = document.createElement('script');
+      script.src = config.google;
+      script.id = 'google-api';
+
+      script.onload = () => {
+        const gapi = window.gapi;
```
shaunanoordin: I take it that it's safe to assume that when the `<script>` loads, we can be guaranteed that `window.gapi` is there. At the moment I can't think of any scenario where the script loads but .gapi is removed or invalid, so that's 👌 , but let me know if you can spot otherwise.

src/components/common/SuperDownloadButton.jsx
```diff
@@ -61,12 +61,12 @@ class SuperDownloadButton extends React.Component {
     // Grommet Button determines disabled state based on presence of onClick handler...
     return (
       <Button
-        className={this.props.className}
+        className={this.props.className || null}
```
shaunanoordin: Out of curiosity, what was happening here? When className was `undefined`, it printed the class name as the string `"undefined"` ?

src/ducks/caesar-exports.js
```diff
@@ -79,6 +80,34 @@ Effect('getCaesarExport', (data) => {
     });
 });
 
+Effect('exportToGoogleDrive', (csv) => {
```
shaunanoordin: Whoah, this part went a bit eldritch. Minor request, but can you please add a link (either as a Github comment or a comment in-code) to the Google API documentation for this part?

That way if someone needs to debug this in the future, they'll be able to find the references that you were using. (Sanity disclaimer: that said, I'm aware links will change/get outdated, but at least it'll be something to go by.)

src/components/common/SuperDownloadButton.jsx
srallen: Oh, I was trying to prevent that from happening. It should be using null if this.props.className is falsey. This isn't doing what I thought it would.

src/ducks/caesar-exports.js
srallen: Sure, can do. Though the docs aren't fantastic. Between the docs and a helpful stackoverflow example is how I got this working.

src/components/astro/GoogleDriveExportButton.jsx
srallen: Yeah, I think the only scenario that I can think of is if the gapi client is deprecated and if it is, then it will probably fail to load anyway.

srallen:
@shaunanoordin I've added comments about the Google Drive API and I'll leave a comment in #26 about using the actual caesar export CSV instead of a test blob.

shaunanoordin:
Sweeeet 👍  Now let's get this PR merged!

Issue #72, opened 2017-10-16
Issues creating I2A Classrooms


shaunanoordin:
I just caught this while I was checking out PR #69 (tee hee) but I can't be sure: when this issue was introduced, whether it's a change on the app code or the Edu API server, or whether it's a temporary (service) issue on the staging server.

## Issue
When creating an I2A Classroom (on localhost+OSX+Chrome as `zootester1`), two unexpected things happen:
- The Create Classroom modal doesn't close
- The three I2A Assignments aren't auto-created

![screen shot 2017-10-16 at 14 32 16](https://user-images.githubusercontent.com/13952701/31614947-e43fcd34-b27f-11e7-93ad-03f476f585ae.png)

### Observations
- Process is as follows:
  1. sign in as `zootester1` on localhost, then go to I2A
  2. Click "Create" (Classroom), then fill in a name for the classroom (`zooterter1 test - 16 Oct 2017`), the click OK ("Create").
  3a. Expected: modal should close. Classroom Manager should show newly created classroom with 3 I2A Assignments.
  3b. Actual: modal doesn't close. Error message appears: "Something went wrong". **Classroom is created,** but it has no pre-generated I2A Assignments.
- In EduAPI, `/assignments?classroom_id={RECENTLY_CREATED_CLASSROOM}` seems to issue a 500. `/classrooms` and `/assignments?classroom_id={OTHER_CLASSROOM}` seem to be OK though.
- The modal not closing may be _unrelated_ to the issue with the `/assignment` endpoint.

### Status
Investigating

srallen:
The payload sent with the assignment creation needs updating since the PR for programs got merged. It no longer requires the program id, since the classroom should be associated with the program. I started to work on updating this, but uncovered a minor bug with the API that should be fixed in the incoming zooniverse/education-api#70.

Thanks for documenting this as it is a good reminder that this is still needing fixing.

srallen:
I believe this was fixed between zooniverse/education-api#70 and #73

Issue #54, opened 2017-09-14
Create classroom with a program id link


srallen:

srallen:
Closed by #73

PR #73, opened 2017-10-16
Classroom program link

srallen:
For #72 and #54, this PR refactors the methods for creating a classroom and auto-creating the assignments for I2A classrooms. 
- Classroom creation now includes the program id in its relationship. This is to support filtering classrooms by program
- Auto-create assignments no longer needs the program id and the workflow id is sent in the attributes

This isn't fully working yet because of an API bug that is fixed in zooniverse/education-api#70 still needs to merge.

### DIFF
--- a/src/containers/classrooms/ClassroomFormContainer.jsx
+++ b/src/containers/classrooms/ClassroomFormContainer.jsx
@@ -95,8 +95,8 @@ export class ClassroomFormContainer extends React.Component {
       Actions.createAssignment(assignmentData);
     })).then(() => {
       Actions.classrooms.toggleFormVisibility();
-      Actions.getClassroomsAndAssignments();
-    })
+      Actions.getClassroomsAndAssignments(this.props.selectedProgram);
+    });
   }
 
   render() {
--- a/src/containers/classrooms/ClassroomFormContainer.jsx
+++ b/src/containers/classrooms/ClassroomFormContainer.jsx
@@ -38,14 +38,12 @@ export class ClassroomFormContainer extends React.Component {
 
   createClassroom() {
     const classroomData = {
-      data: {
-        attributes: this.props.formFields,
-        relationships: {
-          program: {
-            data: {
-              id: this.props.selectedProgram.id,
-              type: 'programs'
-            }
+      attributes: this.props.formFields,
+      relationships: {
+        program: {
+          data: {
+            id: this.props.selectedProgram.id,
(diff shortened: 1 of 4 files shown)


shaunanoordin:
Update on the external EduAPI dependency:
- https://github.com/zooniverse/education-api/pull/70 has been merged and deployed.
- https://github.com/zooniverse/education-api/pull/71 has been merged (can't confirm deploy, though)

As of the time of this update, **I2A Classrooms can no longer be created,** but no error messages are present - the Classroom Manager simply does not list the recently created Classroom, even after a page refresh.

For example, I'd send this:
```
POST https://education-api-staging.zooniverse.org/teachers/classrooms/
Body:
{  
  "data":{  
    "attributes":{  
      "name":"zootester1 - 2017.10.18 14:00",
      "subject":"",
      "school":"",
      "description":""
    },
    "relationships":{  
      "program":{  
        "data":{  
          "id":"1",
          "type":"programs"
        }
      }
    }
  }
}
RESPONSE: 201 Created
```

...and then, when I run `GET /classrooms`, the response data will not contain the classroom "zootester1 - 2017.10.18 14:00".

Tested on localhost+OSX+Chrome while logged in as `zootester1`

At this point, I'm unsure if the Classroom simply isn't registered properly within the EduAPI database, (perhaps it's... missing Program information? Maybe?) or if it is registered but just isn't returned in the GET response. (Maybe it no longer 'belongs' to zootester1, somehow?)

srallen:
@shaunanoordin you can check that API is on the latest deployed commit by going to https://education-api-staging.zooniverse.org/ in the browser. I think @zwolf added this a few weeks ago.

I'll check out the behavior in a few minutes and will confirm if I see the same thing.

srallen:
@shaunanoordin I can confirm I'm seeing the same behavior. @zwolf perhaps there are more issues with Active Integration?

I just POSTed to create a classroom as zootester1, see a 201 response, with a classroom with id `315`, but the subsequent classrooms GET request does not return the newly created classroom.

zwolf:
Classrooms are being successfully created, but GET requests to `/teachers/classrooms` should include the program ID as a query param in order to retrieve the ones you are creating. 

It defaults to nil, which does currently return some classrooms because presently on staging there are classrooms that have no program association. This will not be the case in the future and I can have it return an error if program_id is missing if that is clearer.

Whether that request, without

PR #74, opened 2017-10-19
Darien classrooms table

srallen:
This will have to merge after #73 because it is using the program filter for classrooms, so please ignore the changes that pertain to #73 until this is rebased.

This adds a view for the Darien educators page which is basically the classrooms manager which toggles between the table list view or the classroom editor. Darien's table will be slightly different than I2A's:
- the ability to create and edit assignments which I imagine will open a modal with a form like classroom create/edit does. Right now I just have placeholder buttons for creating and editing an assignment in the markup.
- The actual data in the table for the assignments. I replaced the export data column for the due date. 

I _really_ wanted to try to be clever with determining which classroom table to display by trying out some higher order components, but after puzzling on this for a while I ended up settling with a ternary. The nest of components to get down to the classrooms table is a bit convoluted and I'd really like to try to clean this up at some point as I'm finding it hard to follow myself and I wrote most of it. 

Part of the issue is React Router 4's new way of declarative routes, so there are a few components who's primary purpose is to contain the `<Switch />` and `<Routes />` for whatever nested routes they are.

The other issue is that Darien and I2A share all of the classroom components up until the table, so I couldn't figure out a clever way to have the routes programmatically determine the table since the table itself isn't going to be a route; it's just a child component of the `<ClassroomsManagerContainer />` component that is assigned to the `/educators/classrooms` route. The  I2A `/educators` also doesn't exist on its own; the I2A home page redirects to the educators page since they won't have a student interface. 

If you have any good ideas on how to make this better, I'm definitely open to it.

### DIFF
--- a/src/containers/classrooms/ClassroomsManagerContainer.jsx
+++ b/src/containers/classrooms/ClassroomsManagerContainer.jsx
@@ -11,19 +11,6 @@ import {
 } from '../../ducks/programs';
 
 export class ClassroomsManagerContainer extends React.Component {
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      classroomToDelete: null,
-      showConfirmationDialog: false
-    };
-
-    this.closeConfirmationDialog = this.closeConfirmationDialog.bind(this);
-    this.deleteClassroom = this.deleteClassroom.bind(this);
-    this.maybeDeleteClassroom = this.maybeDeleteClassroom.bind(this);
-  }
-
   componentDidMount() {
     if (this.props.selectedProgram) Actions.getClassroomsAndAssignments(this.props.selectedProgram);
   }
@@ -38,41 +25,13 @@ export class ClassroomsManagerContainer extends React.Component {
     Actions.classrooms.setClassrooms(CLASSROOMS_INITIAL_STATE.classrooms);
   }
 
-  maybeDeleteClassroom(id) {
-    this.setState({ classroomToDelete: id, showConfirmationDialog: true });
-  }
-
-  closeConfirmationDialog() {
-    this.setState({ classroomToDelete: null, showConfirmationDialog: false });
-  }
-
-  deleteClassroom() {
-    if (this.state.classroomToDelete === null) return;
-
-    Actions.deleteClassroom(this.state.classroomToDelete).then((response) => {
-      // TODO: For API optimization, do we want to instead manually remove the classroom
-      // out of local app state instead of making another API call
-      Actions.getClassroomsAndAssignments(this.props.selectedProgram);
-      this.closeConfirmationDialog();
-
-      if (response) {
-        Actions.classrooms.setToastState({ status: 'ok', message: 'Classroom deleted' });
-      }
-    });
-  }
-
   render() {
     return (
       <ClassroomsManager
         classrooms={this.props.classrooms}
         classroomInstructions={this.props.classroomInstructions}
         classroomsStatus={this.props.classroomsStatus}
-        classroomToDelete={this.state.classroomToDelete}
-        closeConfirmationDialog={this.closeConfirmationDialog}
-        deleteClassroom={this.deleteClassroom}
         match={this.props.match}
-        maybeDeleteClassroom={this.maybeDeleteClassroom}
-        showConfirmationDialog={this.state.showConfirmationDialog}
         showForm={this.props.showForm}
       />
     );
@@ -95,7 +54,6 @@ const mapStateToProps = (state) => ({
   programs: state.programs.programs,
   selectedClassroom: state.classrooms.selectedClassroom,
   selectedProgram: state.programs.selectedProgram,
-  showConfirmationDialog: state.classrooms.showConfirmationDialog,
   showForm: state.classrooms.showForm
 });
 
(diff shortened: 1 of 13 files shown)


srallen:
Noting that this is for issue #70

shaunanoordin:
I've done a quick review on the user side of things, and here are some observations - although much of what I say may be rendered moot when this PR is rebased.

- Overall layout & presentation: 👍 
  - Also 👍  on the placeholders that will eventually open the Create Assignment modal; that's the path I'd like to go down.
  - Thought: How the Create Assignment modal will be built, though, will need better planning, since Darien's Create Assignment will have a project-specific feature that allows it to communicate with the Map Explorer. (i.e. how can this be generalised?)
- Quirks:
  - [medium] I can create Classrooms in Darien, but the Classroom won't be listed until I refresh the page.
    - This works properly in I2A, though.
    - Thought: is there some code in I2A that refreshes the Classrooms Manager _only after_ Assignments are auto-created? If so, that may explain why Darien has that quirk.
  - [major] Delete Classroom button doesn't work on Darien. (EDIT: I2A too)

srallen:
This needs a rebase now and delete classroom button sounds like a regression, so I'll work on patching that up. I'll take a look at the request for refreshing the classrooms list too for Darien.

src/containers/classrooms/ClassroomsTableContainer.jsx
```diff
@@ -1,82 +1,91 @@
...
+      if (response) {
+        Actions.classrooms.setToastState({ status: 'ok', message: 'Classroom deleted' });
+      }
+    });
   }
 
-  selectClassroom(classroom) {
-    Actions.classrooms.selectClassroom(classroom);
+  determineClassroomsTable() {
+    const ClassroomsTable = this.props.selectedProgram.custom ? DarienClassroomsTable : AstroClassroomsTableContainer;
```
shaunanoordin: OK, now I see what you meant when you said _"...after puzzling on this for a while I ended up settling with a ternary"_

I think this is the way to go, at least for now. Re: _"I'd really like to try to clean this up at some point",_ I'm all for it, but I'm OK to launch something that works first and refactor later. That said...

Thought: I imagine the ideal solution would be to properly plan out a kind of "generic Education Program" system similar to Panoptes' "generic Build-Your-Own Project" system, then plunk the switching code into a central "ProgramsController"/"ProgramsConfigToComponentTranslator" thingy...

```
switch (this.props.selectedProgram.settings.programType) {
  case "wildlife_camera_trap_classroom": ClassroomsTable = DarienClassroomTable; break;
  case "standard_auto_created_assignments_classroom": ClassroomsTable = AstroClassroomsTableContainer; break;
  ...
}
return ClassroomsTable;
```

...but that _might_ be a pretty darn deep rabbit hole to go down.

src/containers/classrooms/ClassroomsTableContainer.jsx
```diff
@@ -1,82 +1,91 @@
...
+    return (
+      <ClassroomsTable
+        assignments={this.props.assignments}
+        assignmentsStatus={this.props.assignmentsStatus}
+        classrooms={this.props.classrooms}
+        match={this.props.match}
+        maybeDeleteClassroom={this.props.maybeDeleteClassroom}
```
shaunanoordin: Unsure: Is this missing `deleteClassroom={this.props.deleteClassroom}` ?

src/containers/classrooms/ClassroomsManagerContainer.jsx
```diff
@@ -6,84 +6,54 @@ import ClassroomsManager from '../../components/classrooms/ClassroomsManager';
...
-
   render() {
     return (
       <ClassroomsManager
         classrooms={this.props.classrooms}
         classroomInstructions={this.props.classroomInstructions}
         classroomsStatus={this.props.classroomsStatus}
-        classroomToDelete={this.state.classroomToDelete}
-        closeConfirmationDialog={this.closeConfirmationDialog}
-        deleteClassroom={this.deleteClassroom}
```
shaunanoordin: Nyerp this one's bothering me, `deleteClassroom` is removed _here,_ but I'm assuming it should have been added back/moved to _somewhere else._ I still can't tell where it's supposed to go, though.

srallen: That's right. I shuffled the classroom delete methods into `ClassroomsTableContainer`

src/containers/classrooms/ClassroomsTableContainer.jsx
srallen: `maybeDeleteClassroom={this.maybeDeleteClassroom}` This should open the dialog and `this.deleteClassroom` is the click handler for confirmation. I missed removing the `this.props` part here.

src/containers/classrooms/ClassroomsManagerContainer.jsx
srallen: It was my attempt at making sense of component organization and what should be responsible for what. Clearly, it's all still hard to follow 🤷‍♀️

srallen:
I just rebased, so this should be a bit easier to see the changes specific to this PR.

src/containers/classrooms/ClassroomsTableContainer.jsx
srallen: It's a good idea, but not worth the time right now even though I dislike what I ended up doing! If another curricula/educational program is added after I2A and the Wildcams, then that's when this should be refactored.

shaunanoordin: 👍

srallen:
@shaunanoordin those two issues should now be fixed. I found a bug with classroom creation, too, but I'll save that for another PR.

PR #75, opened 2017-10-24
Fix classroom form state

srallen:
This does two things:

- Fixes a possible regression I introduced that I noticed while debugging #74 
  - The classroom create/edit form wasn't clearing out the form fields on unmount, so when you created a classroom and then clicked to create another (or edit), the form field had the previous content.
- This also adds a consistent sort order for the I2A assignments

### DIFF
--- a/src/ducks/classrooms.js
+++ b/src/ducks/classrooms.js
@@ -141,7 +141,7 @@ Effect('getClassroomsAndAssignments', (selectedProgram) => {
         // loop through the number of pages to request all of the data
         // and concatenate the response data together for the app state
         // Neither Pagination nor infinite scroll would be good UX for current table design.
-        Actions.getAssignments(classroom.id);
+        Actions.getAssignments({ classroomId: classroom.id, selectedProgram });
       });
     }
   });
--- a/src/ducks/classrooms.js
+++ b/src/ducks/classrooms.js
@@ -92,11 +92,7 @@ const updateFormFields = (state, formFields) => {
 Effect('getClassrooms', (selectedProgramId) => {
   Actions.classrooms.setStatus(CLASSROOMS_STATUS.FETCHING);
 
-<<<<<<< a1176a69c771beaef9d5be1ab7b99fc292f79de9
   return get('/teachers/classrooms/', [{ program_id: selectedProgramId }])
-=======
-  return get('/teachers/classrooms/', [{ 'filter[programs]': selectedProgramId }])
->>>>>>> Add program filter
     .then((response) => {
       if (!response) { throw 'ERROR (ducks/classrooms/getClassrooms): No response'; }
       if (response.ok &&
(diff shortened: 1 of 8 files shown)


srallen:
Note: I'm not really sure why all those commits are showing up... It looks like something odd happened while rebasing. I can squash those if you want me too... or we can squash on merge.

src/ducks/classrooms.js
```diff
@@ -141,7 +141,7 @@ Effect('getClassroomsAndAssignments', (selectedProgram) => {
         // loop through the number of pages to request all of the data
         // and concatenate the response data together for the app state
         // Neither Pagination nor infinite scroll would be good UX for current table design.
-        Actions.getAssignments(classroom.id);
+        Actions.getAssignments({ classroomId: classroom.id, selectedProgram });
```
shaunanoordin: Out of curiosity, is there any reason the function `getAssignments()` is structured like `getAssignments(data)` where `data = {classroomId, selectedProgram}` instead of `getAssignments(classroomId, selectedProgram)`?

It's functionally the same, but from a readability standpoint, it's not very clear what arguments/values `getAssignments()` is expecting.

The only advantages that I can see for using a single 'compound variable' instead of spelling out the different variables are if:
- the compound variable (`data`) supports different kinds of key-values (e.g. an advanced search query, `data = {color:'red', species:'dragon', optionalAge: null }` ), or
- the `data` is structured in a specific way that doesn't matter to the function(), but matters to the API (e.g. createAssignnments())

srallen: Jumpstate's Effect either wasn't designed to work or had a bug when I tried to pass it two separate arguments! 😢

shaunanoordin:
At the moment, I'm encountering an error loading I2A Assignments:
screen shot 2017-10-24 at 20 52 21

`master` looks OK, so it seems to be some issue with this bit here...
```
props.assignments[classroom.id].map((assignment) => {
  const projectUrl = (assignmentsMetadata && assignmentsMetadata[assignment.workflowId]) ?
    `${config.zooniverse}/projects/${assignmentsMetadata[assignment.workflowId].slug}/classify?group=${classroom.attributes.zooniverse_group_id}` :
    null;
```
...where, for some reason, there are cases where assignment is null/undefined. (Specifically, `props.assignments['316'] = [undefined, undefined, undefined]` - whuuuh?)

src/ducks/classrooms.js
shaunanoordin: Oh gods, that's right! I totally forgot about Jumpstate's "one Effect, one argument" rule. 🤦‍♂️

srallen:
@shaunanoordin thanks for catching that. There is an edge case of I2A classrooms that do not have assignments. These may have been created while there was that issue with the API with linking assignments correctly. zootester1 has one of these classrooms. I moved the code to sort the assignments out into its own function and do a existential check on the assignments first before returning the sorted assignments.

PR #76, opened 2017-10-30
Add darien assignment form

srallen:
These sets up the scaffold for a working create and edit assignment form for WD:
- Updates grommet dependency to 1.8.3
- Adds the `AssignmentFormDialog`, `AssignmentFormContainer` and `AssignmentForm` components
- Adds some scaffolding for the delete action for assignments
- I deleted a copy of `AstroClassroomsTableContainer` that had a typo in its file name, so there were basically two copies of the same component.
- I fixed a slight typo variation in the i2A program metadata, where I'm storing how to create the 3 standard assignments they use.

**Note that assignment creation still does not work yet.** It's known it will 422. I don't have the data structured correctly yet. However, I wanted to get this reviewed before this branch got too much bigger!

Still left to do in a follow up PRs:
- A working POST for assignment create
- A working list of students to add
- A working subject viewer
- A working delete assignment action


src/components/assignments/AssignmentForm.jsx
```diff
@@ -0,0 +1,107 @@
...
+      </Footer>
+    </Form>
+  );
+};
+
+AssignmentForm.defaultProps = {
+  onChange: () => {},
+  fields: {
+    name: '',
+    subject: '',
```
shaunanoordin: I'm not sure if this is supposed to be `fields.subject` and `fields.school`; I think we need default props for `fields.name|description|duedate|classifications_target`

src/components/assignments/AssignmentForm.jsx
```diff
@@ -0,0 +1,107 @@
...
+  heading: '',
+  onSubmit: () => {},
+  students: [],
+  submitLabel: 'Submit'
+};
+
+AssignmentForm.propTypes = {
+  onChange: PropTypes.func,
+  fields: PropTypes.shape({
+    classificationsTarget: PropTypes.number,
```
shaunanoordin: The `AssignmentForm` component uses `props.fields.classifications_target` instead of `props.fields.classificationsTarget`; I need to check any parent component using `AssignmentForm` to ensure consistency.

shaunanoordin: 👌  OK, confirmed: every component uses `classifications_target`.

src/containers/assignments/AssignmentFormContainer.jsx
```diff
@@ -0,0 +1,123 @@
...
+    this.onSubmit = this.onSubmit.bind(this);
+  }
+
+  componentWillUnmount() {
+    this.resetFormFields();
+  }
+
+  onChange(event) {
+    if (event.target && event.target.id && event.target.value) {
+      const fields = { ...this.props.formFields, [event.target.id]: event.target.value };
```
shaunanoordin: Not a review comment, but a personal one: _ooh!_  TIL how to use `{ [key_name]: value }` to create on-the-fly keys for {objects}, and that's kinda amazing.

src/containers/assignments/AssignmentFormContainer.jsx
```diff
@@ -0,0 +1,123 @@
...
+      });
+    }
+  }
+
+  createAssignment() {
+    const assignmentData = {
+      attributes: {
+        name: this.props.formFields.name,
+        metadata: {
+          classification_target: this.props.formFields.classification_target,
```
shaunanoordin: It seems this is using `classification_target` instead of the plural form `classification_target`; I'm assuming that the plural form is the correct one.

Side note: I noticed the non-plural spelling on `ClassroomFormContainer.jsx` as well.

src/components/assignments/AssignmentForm.jsx
```diff
@@ -0,0 +1,107 @@
...
+            value={props.fields.duedate}
+          />
+        </FormField>
+        <FormField htmlFor="classifications_target" label="Number of images per student" help="required">
+          <NumberInput
+            id="classifications_target"
+            min={1}
+            onChange={props.onChange}
+            required={true}
+            value={props.fields.classifications_target}
```
shaunanoordin: I should make a note here before I forget to check:

The `classifications_target` value needs to be a string, correct? If so, would `<NumberInput>` submit the field's values to the `onChange()` function as a string or a number?

src/containers/classrooms/ClassroomsTableContainer.jsx
```diff
@@ -34,17 +39,40 @@ class ClassroomsTableContainer extends React.Component {
...
+      classroomToDelete: null,
+      showConfirmationDialog: {
+        assignment: false,
+        classroom: false
+      }
+    });
+  }
+
+  maybeDeleteAssignment(id) {
+    this.setState({ assignmentToDelete: id, showConfirmationDialog: { assignment: true } });
```
shaunanoordin: This `setState()` call appears to 'erase' the `state.showConfirmationDialog.classroom` key-value; i.e. replacing `showConfirmationDialog: { assignment: false, classroom: false }` with `showConfirmationDialog: { assignment: true }` .

That said, in practice I detect no issues with this, as `undefined` is functionally similar enough to `false` that there shouldn't be any hiccups.

shaunanoordin:
### Side Note: Analysing the "Create Assignment" 422 Error
With this PR, when I choose to Create an Assignment via the WildCam Darien page, I'm sending data that looks like this:

```
{  
  "data":{  
    "attributes":{  
      "name":"Test 1",
      "metadata":{  
        "description":"test",
        "duedate":"11/13/2017 3:00 pm"
      }
    },
    "relationships":{  
      "classroom":{  
        "data":{  
          "id":"321",
          "type":"classrooms"
        }
      }
    }
  }
}
```

Now if I compare it to a Create Assignment action in WildCam Gorongosa Lab, I'll see...
```
{  
  "data":{  
    "attributes":{  
      "name":"TEST",
      "metadata":{  
        "classifications_target":"10",
        "description":"Simone is a member.",
        "duedate":"2017-01-01",
        "filters":{  
          "species":[  
            "civet"
          ]
        },
        "subjects":[...a bunch of string IDs...]
      }
    },
    "relationships":{  
      "classroom":{  
        "data":{  
          "id":"361",
          "type":"classrooms"
        }
      },
      "student_users":{  
        "data":[...a bunch of id-type objects...]
      },
      "subjects":{  
        "data":[...a bunch of id-type objects...]
      }
    }
  }
}
```

While it's worth double-checking that it's possible (IIRC, _it should be_ ) to create an Assignment with 0 assigned Students and 0 assigned Subjects, let's ignore `relationships` for now and focus on the differences in `attributes`. We notice that...

1. `metadata.classifications_target` is missing from the Darien Create Assignment call; this may be related to the `classifications_target` confusion issue above. IIRC, classifications_target is a _required_ parameter, so missing this may be causing the 422.
2. `metadata.duedate` uses a complete different time-date format; for one, the WildCam Darien call actually includes a _time._
  - Side note: the YYYY-MM-DD format is probably the best format to use to avoid confusion between UK and US date formats.
3. I'm not sure if `metadata.filters` is a required parameter; I know I can submit a call with no filters, but that may be sent as an empty object `{}` . Again, I need to double check this.

src/components/assignments/AssignmentForm.jsx
srallen: Yep, probably cruft from when I copied and pasted from the classroom's form as a place to start.

src/components/assignments/AssignmentForm.jsx
srallen: Yep, and the PropType here is wrong too.

src/containers/assignments/AssignmentFormContainer.jsx
srallen: One of the great things ES6 gave us!

src/containers/assignments/AssignmentFormContainer.jsx
srallen: I made the singular v. plural mistake all over the place and I don't think I caught until after I started working on this. It might be the source of the 422?

src/components/assignments/AssignmentForm.jsx
srallen: Yeah, this one kind of boggles me. I figured out this should be a string after I begun this and it doesn't make sense, especially when it could be used to calculate stats like I'm trying to do in #90. Why is this a string instead of a number for the API?

srallen: FWIW, I'll make it a string since that's what is being used with the API, I just find it strange.

src/containers/classrooms/ClassroomsTableContainer.jsx
srallen: doh, gotta remember that setState doesn't deep merge.

srallen:
I'm working on the 422. To follow up on your comments:
1. I've corrected all of singular v. plural mistakes with `classifications_target`
2. I've standardized the datetime format using a format prop Grommet offers. They default to `M/D/YYYY h:mm a`, however, `YYYY-MM-DD` can be used.
3. I doubt the filters are required, but I'm not sure either. This form will have to be able to submit filters regardless. 

However, the API response gives us a clue. Nested in the error object we get this detail: `"has an invalid nested value ("workflow_id" => nil)"` I think this was one of the changes necessary for I2A, so we'll need to get the docs updated for this. I think sending a `workflow_id` is required now. For Wildcam, this may be the workflow id that is going to be cloned? I'll confirm with @zwolf .

shaunanoordin:
## PR Review Update
Nice! Thanks @srallen , the recent changes address the pertinent issues for getting this PR merged. 👍 

### Status
LGTM, this PR is now getting merged.

PR #78, opened 2017-11-07
Use program mock as fallback

srallen:
This updates the error handler for requesting the programs. If it 404s, like it currently does in production since the back end API hasn't been deployed yet and still has no notion of a program, then it'll fallback to the programs mocks.

### DIFF
--- a/src/ducks/programs.js
+++ b/src/ducks/programs.js
@@ -193,7 +193,14 @@ Effect('getPrograms', () => {
         return response.body.data;
       }
       throw 'ERROR (ducks/programs/getPrograms): Invalid response';
-    }).catch(error => handleError(error));
+    }).catch((error) => {
+      if (error.status !== 404) handleError(error)
+      if (error.status === 404) {
+        // Fallback to showing the mocks
+        Actions.programs.setStatus(PROGRAMS_STATUS.SUCCESS);
+        Actions.programs.setPrograms(programsArray);
+      }
+    });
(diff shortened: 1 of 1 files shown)


src/ducks/programs.js
```diff
@@ -193,7 +193,14 @@ Effect('getPrograms', () => {
         return response.body.data;
       }
       throw 'ERROR (ducks/programs/getPrograms): Invalid response';
-    }).catch(error => handleError(error));
+    }).catch((error) => {
+      if (error.status !== 404) handleError(error)
```
shaunanoordin: [minor] missing semicolon;

shaunanoordin:
The code changes look good, but I have to ask if quietly falling back to the `programMocks` is going to be a permanent addition to the code, or if it's just a temporary solution to get production running.

The reason I ask is that it may make debugging particularly difficult, if - on production - we can't really be sure if we're looking at the actual Program data or the mock data. If this is going to be any sort of long-term solution, can I request that we at least add in something like `console.log('Nyerp, looks like the API has no Program data, so FYI we\'re falling back to some hardcoded defaults');` ?

srallen:
I'm intending this to be temporary. I'd like to remove it once the production API is deployed. Maybe even before then. It seems like a good idea to put a comment about that regardless, so let me do that before merging.

shaunanoordin:
LGTM! 👍

PR #79, opened 2017-11-08
Remove intro assignment

srallen:
It's been decided that the introductory assignment wasn't a good fit for use of the Zooniverse's project builder and platform. This PR removes it. WIP because staging API is on the fritz at the moment and the programs in the database still need updating. I also fix a typo and make the classifications_target a string since that's what they actually are on assignments. I fixed this in #76, but this is a better fit for updating this. I'll rebase #76 after this merges.

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -35,7 +35,7 @@ const Main = ({ admin, location }) => {
   }
 
   return (
-    <App centered={false} className="app-layout" inline={true}>
+    <App centered={false} className="app-layout">
       {admin &&
         <AdminLayoutIndicator />}
       <Box>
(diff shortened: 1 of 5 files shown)


srallen:
A couple more notes on changes:
- I removed `inline={true}` from the main app component to work around a grommet bug I found with the Layer component. I've informed the team of the bug, but haven't opened a issue for them because I haven't successfully recreated it in a codepen for them. And other things are taking higher priority
- I added a timeout in the call to get assignments because i2a was sometimes getting the classroom and getting either one assignment or a null response for the newly created classroom. The timeout is a quick and dirty fix for this.

srallen:
Gonna go ahead and merge since I'd like to get this in before some other branches I'm working on.

Issue #67, opened 2017-10-10
Actual stats for students


srallen:
Replace the placeholder numbers with actual completion stats for the students in the teacher's classroom view.

PR #88, opened 2017-11-10
Real stats prep

srallen:
For #67, in anticipation of using real stats for the student's progress, this PR
- Extracts `blobbifyData` and `generateFileName` into its own helper js file to be shared between Darien and I2A
- Removes the Introduction assignment from the table since that won't be a Zooniverse project now
- Makes the download csv file name prefix a prop for the SuperDownloadButton
- Fixes the TypeError seen on the classroom edit page when assignments were requested
- Fixes linter errors in SuperDownloadButton

Question about the SuperDownloadButton:
We officially only support the current version and last previous version of browser. For Safari, that's now version 10 and 11. Do we still need the work around written in the SuperDownloadButton for Safari?

### DIFF
--- a/src/components/astro/ExportModal.jsx
+++ b/src/components/astro/ExportModal.jsx
@@ -58,9 +58,10 @@ const ExportModal = ({ caesarExport, caesarExportStatus, onClose, showModal }) =
           <Box direction="row">
             <SuperDownloadButton
               className="export-modal__button"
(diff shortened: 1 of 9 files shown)


srallen:
I'm gonna go ahead and merge this as I want to continue forward progress on finishing the stats feature including the grade export button.

PR #90, opened 2017-11-13
Real stats

### LINKED ISSUE #67: Actual stats for students

srallen:
Closes #67. This should be merged after #88. This adds the math for overall classroom completeness as well as the math for individual student progress.

To do after this PR:
- Add toggle to switch between percentage and actual number view
- Finish the export grades class method
- Manually test that the API is ingesting the classifications and counting.


src/components/astro/AstroClassroomsTable.jsx
```diff
@@ -21,8 +21,15 @@ import {
   CLASSROOMS_INITIAL_STATE, CLASSROOMS_PROPTYPES
 } from '../../ducks/classrooms';
 
-const AstroClassroomsTable = (props) => {
-  // TODO: Replace classroom.zooniverse_user_group_id with actual serialized user group id when available
+function calculateCompleteness(assignment, classroom) {
+  const classificationsTarget = +assignment.metadata.classifications_target;
```
shaunanoordin: Two questions:
1. Is there any chance that `classificationsTarget` can ever be 0?
2. I assume the `+` is a method of converting what could possibly be a string into a number, correct?

src/components/astro/AstroClassroomsTable.jsx
```diff
@@ -21,8 +21,15 @@ import {
...
+  const numberOfStudents = assignment.studentAssignments.length;
+  const numberOfCompletedClassifications = classroom.classificationsCount;
+  if (numberOfStudents === 0) return 0;
```
shaunanoordin: Actually, all things considered, might it be a good idea in any case to change this to `if (numberOfStudents === 0 || classificationsTarget === 0) ...` to accommodate for those weird edge case Assignments? I'm not sure if the `assignment.classifications_target must be > 0` rule is enforced anywhere asides from the front end Assignment creation system.

(Or for the utterly paranoid, `if (!(numberOfStudents >= 1) || !(classificationsTarget >= 1) ...` )

src/components/classrooms/ClassroomEditor.jsx
```diff
@@ -175,15 +176,26 @@ const ClassroomEditor = (props) => {
...
             props.assignments[props.selectedClassroom.id].length > 0 &&
-              students.map(student =>
-                (<TableRow className="manager-table__row-data" key={`classroom-student-${student.id}`}>
+              students.map((student) => {
+                const galaxyAssignment = props.assignments[props.selectedClassroom.id].filter(
+                  assignment => assignment.name === i2aAssignmentNames.first);
+                const hubbleAssignment = props.assignments[props.selectedClassroom.id].filter(
+                  assignment => assignment.name === i2aAssignmentNames.second);
+
+                // Why are the ids in the student_user_id property numbers?!?!?!
```
shaunanoordin: ??!??

I... maybe... API magic??

src/components/classrooms/ClassroomEditor.jsx
```diff
@@ -175,15 +176,26 @@ const ClassroomEditor = (props) => {
...
+
+                return (<TableRow className="manager-table__row-data" key={`classroom-student-${student.id}`}>
                   <td headers="student-name">
                     {(student.zooniverseDisplayName && student.zooniverseDisplayName.length > 0)
                       ? <span>{student.zooniverseDisplayName}</span>
                       : <span className="secondary">{student.zooniverseLogin}</span> }
                   </td>
-                  <td headers="assignment-galaxy">{Math.floor(Math.random() * 100)}%</td>
-                  <td headers="assignment-hubble">{Math.floor(Math.random() * 100)}%</td>
+                  <td headers="assignment-galaxy">{(galaxyStudentData[0].attributes.classifications_count / (+galaxyAssignment[0].metadata.classifications_target)) * 100}%</td>
```
shaunanoordin: 1. Is `attributes.classifications_count` always a number, not a string?
  - (Also note conversation with Amy on Slack on #education-api at 11am 15 Nov 2017 Chicago, where some comments were made about converting numbers to string as a standard... I'm not sure if there's going to be some blanket update to the number values that may affect this.)
2. I was going to ask if `classifications_target` would ever be 0, but then I realised that... maybe we don't need to worry too much, since with JS, if you want to display some nonsensical value like `123 / 0` or `123 / undefined`, the code will just shrug and display NaN or Infinity, instead of crashing. I'll doublecheck this, but if this interpretation is correct, then it's _safe enough_ to display data like this without having to worry about the divide-by-0 cases.

src/components/astro/AstroClassroomsTable.jsx
shaunanoordin: EDIT: I retract this comment, as this is just being too paranoid. If `classificationsTarget` is `0`, the worse that will most likely happen is that `Infinity` is displayed as a value instead of the whole app crashing.

shaunanoordin:
Oh wait, there's one merge conflict that needs to be addressed. I'll take care of this.

shaunanoordin:
👍

Issue #85, opened 2017-11-10
Darien edit classroom view -- shared component reorganization?


srallen:
Darien isn't going to want that same classroom edit page design as I2A because it's still hard coded for the I2A uses. I think I was overly ambitious with trying to make some of these components shared between the two. The programs have distinct view and routing needs, so I think we have a few options:

- Darien gets its own edit classroom page design and we determine which to use similar to what I did in #74. I never really liked this solution
- Or we completely separate the shared components into I2A components and Darien components. This would result in some repetition, but that's not necessarily a bad thing.
- Or we separate the view components entirely and use HOC container components to inject in the class methods we want to share between the functional view components.
- And/or abstract out shared API calls and methods into the appropriate ducks into async actions managed by jumpstate.

PR #94, opened 2017-11-14
Export stats

### LINKED ISSUE #50: Use SuperDownloadButton for grade export
What the issue title says.

srallen:
To be merged after #90. This closes #50. It finishes the export stats class method for the individual classroom edit view. Note: I've renamed the button and method to `Export Stats` and `exportStats` because I think we should avoid saying grades anywhere. As a former IT person who worked for a couple of Universities, even implying we store grades could make us vulnerable to complying with student privacy laws. It's up to the teacher if they want to use these stats in their grading, but they aren't grades.

### DIFF
--- a/src/components/classrooms/ClassroomEditor.jsx
+++ b/src/components/classrooms/ClassroomEditor.jsx
@@ -106,7 +106,12 @@ const ClassroomEditor = (props) => {
             <Button type="button" label="Copy Join Link" />
           </CopyToClipboard>
 
-          <Button type="button" primary={true} label="Export Grades" onClick={props.exportGrades} />
+          <Button
+            type="button"
+            primary={true}
+            label="Export Stats"
+            onClick={props.selectedClassroom && props.selectedClassroom.students.length > 0 ? props.exportStats : null}
+          />
         </Box>
 
         <Box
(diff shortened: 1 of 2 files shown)


srallen:
I've rebase this, so it's ready to be reviewed. Thanks!

src/components/classrooms/ClassroomEditor.jsx
```diff
@@ -177,9 +182,9 @@ const ClassroomEditor = (props) => {
             props.assignments[props.selectedClassroom.id] &&
             props.assignments[props.selectedClassroom.id].length > 0 &&
               students.map((student) => {
-                const galaxyAssignment = props.assignments[props.selectedClassroom.id].filter(
+                const galaxyAssignment = assignments.filter(
```
shaunanoordin: [minor] If you're changing `props.assignments[props.selectedClassroom.id]` to the more sensible `assignments`, wouldn't you also want to update the condition checks as well? i.e.

```
{props.assignmentsStatus === ASSIGNMENTS_STATUS.SUCCESS &&
props.assignments[props.selectedClassroom.id] &&
props.assignments[props.selectedClassroom.id].length > 0 &&
  students.map((student) => {
...
```

shaunanoordin:
WTF? When I'm looking at the I2A classrooms as `zootester1`, I receive a warning in the console that says
> Warning: Failed prop type: The prop `assignments.isRequired` is marked as required in `ClassroomsTableContainer`, but its value is `undefined`.
>    in ClassroomsTableContainer (created by Connect(ClassroomsTableContainer))
but the problem is I can't find anything in our code that asks for `isRequired`

srallen:
@shaunanoordin I have no idea where those warnings are coming from. It's a mystery to me at the moment!

shaunanoordin:
`*shakes fists at the heavens*` Damn you mysterious React warning, NOBODY said assignments is required!

srallen:
> the current ClassroomEditor component might need to be relabelled as AstroClassroomEditor when that happens.
@shaunanoordin this will likely have to happen because as it is now the ClassroomEditor is really for I2A. I've opened #85 to discuss it. I think a mix of pulling more shared methods into the ducks as well as some refactoring should do it. I'm becoming more comfortable with higher order components, so those could help us out possibly too.

PR #95, opened 2017-11-15
Update caesar requests

srallen:
For #26, this is a step forward in getting the caesar requests to work. The 404 we saw before was a red herring. It didn't actually work, for one, because the reduction type data export wasn't available publicly before. This is now fixed and Intro2Astro workflows have public caesar exports.

This PR fixes the POST to create the export, stores the pending export in the redux store and if requested again, checks against the presence of a pending export. If there is a pending export, we request using its id. 

There are a few issues still needing resolving, such as, how do we best check for a pre-existing export so we don't unnecessarily create new exports. Because of the remaining API/UX questions, there are still a few logs and commented out items.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8667,6 +8667,11 @@
         "sugar-client": "1.0.1"
       }
     },
+    "papaparse": {
+      "version": "4.3.6",
+      "resolved": "https://registry.npmjs.org/papaparse/-/papaparse-4.3.6.tgz",
+      "integrity": "sha1-lWbtoOyrE6/LdApiOBxpn0hssUU="
+    },
     "param-case": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/param-case/-/param-case-2.1.1.tgz",
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "jumpstate": "^2.2.2",
     "leaflet": "1.2.0",
     "panoptes-client": "~2.7.2",
+    "papaparse": "~4.3.6",
     "prop-types": "~15.5.10",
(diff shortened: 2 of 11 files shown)


srallen:
Update: Still WIP because some modifications on how Caesar is handling some of the reductions needs changing. I've added papaparse to handle the CSV transformations needed for the I2A data before it gets downloaded.

srallen:
Ok I think this is ready for review:

- This updates the caesar requests to have this kind of request workflow:
  1. If the redux store has a `requestedExport` stored (these are exports that have been requested but are pending), request specifically for that export to check its status. If it's still pending, it will remain as is, if completed, then it is stored as `caesarExport`. If there is an error, then show the error message.
  2. If there are no pending requested Exports, then request for all of the exports for a given subgroup (classroom user group id) and use the most recent export if it is completed. 
  3. If there are no exports at all, then send a request to create one. The response becomes the `requestedExport` in the pending state.

I've written the CSV parsing class method for Hubble's Law, so if there is an export ready and you click to download it, you should get a CSV downloaded that is human readable. This necessitated refactoring the SuperDownloadButton to use a `transformData` method if it is present rather than immediately downloading. I've also added PapaParse to do the in browser csv parsing.

To test this, try to download as zootester1 the hubble's law export for the classroom named as "Classroom Assignments Nov 13."

The pending TO DOs are to write the parsing method for the Galaxy Zoo activity and finish the Google Sheets export option.

srallen:
@shaunanoordin I'm gonna go ahead and merge this since I'm getting far ahead of this and need to focus on finishing up the I2A requirements.

PR #98, opened 2017-11-29
Assignment create

srallen:
This preps the assignment creation for Wildcam Darien including:

- Setting up a staging environment variable
- Adding the hard coded workflow ids and sample subjects to WD and WG program mocks
- Checking for the staging environment and using the appropriate hard coded workflow and sample subject ids if it applies on assignment create
- Updates propTypes where needed

This also includes some fixes for the join link and a deploy to staging npm script. I've already deployed to staging via a different branch because I wanted the I2A instructors to start to test features for me this week on staging. The staging environment update and script were cherry picked on to this branch to come in with this PR. 

Note that the assignment create still doesn't work due to noted permissions issue in the slack channel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "test": "NODE_ENV=development BABEL_ENV=test mocha mocha-setup.js $(find src -name *.spec.jsx) --compilers js:babel-core/register --require mock-local-storage || true",
     "eslint": "eslint .",
     "build": "export BABEL_ENV=production; check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
+    "deploy-staging": "export NODE_ENV=staging; npm run build && publisssh dist zooniverse-static/preview.zooniverse.org/classroom",
     "deploy-production": "export NODE_ENV=production; npm run build && publisssh dist zooniverse-static/classroom.zooniverse.org"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/zooniverse/edu-api-front-end"
   },
   "scripts": {
-    "start": "BABEL_ENV=development check-engines && check-dependencies && webpack-dashboard -p 3999 -- webpack-dev-server --config ./webpack.config.js",
+    "start": "export NODE_ENV=development; BABEL_ENV=development check-engines && check-dependencies && webpack-dashboard -p 3999 -- webpack-dev-server --config ./webpack.config.js",
     "test": "NODE_ENV=development BABEL_ENV=test mocha mocha-setup.js $(find src -name *.spec.jsx) --compilers js:babel-core/register --require mock-local-storage || true",
     "eslint": "eslint .",
     "build": "export BABEL_ENV=production; check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
--- a/src/components/astro/AstroClassroomsTable.jsx
+++ b/src/components/astro/AstroClassroomsTable.jsx
@@ -50,7 +50,7 @@ function AstroClassroomsTable(props) {
           </TableRow>
         </thead>
         {props.classrooms.map((classroom) => {
-          const joinURL = `${window.location.host}/#/${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
+          const joinURL = `${config.origin}/#/${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
           // Can we get linked assignments with classrooms in single get request?
           // No, if we want this, then we need to open an issue with the API
           // TODO replace classifications_target with calculated percentage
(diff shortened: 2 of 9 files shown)


srallen:
Assignment POST is fixed on the API.

package.json
```diff
@@ -11,6 +11,7 @@
     "test": "NODE_ENV=development BABEL_ENV=test mocha mocha-setup.js $(find src -name *.spec.jsx) --compilers js:babel-core/register --require mock-local-storage || true",
     "eslint": "eslint .",
     "build": "export BABEL_ENV=production; check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
+    "deploy-staging": "export NODE_ENV=staging; npm run build && publisssh dist zooniverse-static/preview.zooniverse.org/classroom",
```
shaunanoordin: Aaa, nice! And thank goodness we use hash routing so this preview section actually works. 👌

srallen:
@shaunanoordin Just rebased this just so it's up-to-date. There weren't any merge conflicts.

src/components/astro/AstroClassroomsTable.jsx
```diff
@@ -50,7 +50,7 @@ function AstroClassroomsTable(props) {
           </TableRow>
         </thead>
         {props.classrooms.map((classroom) => {
-          const joinURL = `${window.location.host}/#/${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
+          const joinURL = `${config.origin}/#/${props.selectedProgram.slug}/students/classrooms/${classroom.id}/join?token=${classroom.joinToken}`;
```
shaunanoordin: I'm running into a bit of WTF-ery here. 90% of the time, `config.origin` lists `localhost:3998` as expected but every once in a while I see weirdness like this:

screen shot 2017-12-01 at 18 27 46
_Browser window URL: http://localhost:3998/#/... | Join URL: http://localhost:3998/classroom/#/... | OSX+Chrome+zootester1_

There was a previous case where I'd see the origin listed as `localhost:3998...` instead of `http://localhost:3998...` as well, but as with all matters of WTF-ery, replicating the issue is in itself an issue.

shaunanoordin: _-throws hands up in the air-_ So it looks like my localhost thinks I'm `env=staging` (hence the `/classroom` extension to everything) instead of `env=development`, despite my NODE_ENV being `development`, because... the moon is in the house of Scorpio, who knows. At least I can confine this debug to my machine only and not the code.

srallen: Maybe an explicit `export NODE_ENV=development` needs to be added to line 10 of the start script in the package.json?

shaunanoordin: OH! No, I think it's actually the `webpack.config` - maybe we need to change this line:
```
//webpack.config.js, line 47
new webpack.DefinePlugin({
  'process.env.NODE_ENV': JSON.stringify('staging')
}),
```

shaunanoordin:
Ah, bugger, this one's a head scratcher. OK, so:

**Issue:** `webpack.config.js` needs to be adjusted because the new `config.js` makes a slight (but important) distinction between `env=development` and `env=staging` .

**Option 1:** Currently, `webpack.config.js` sets NODE_ENV to staging.
See line 47:
```
new webpack.DefinePlugin({
  'process.env.NODE_ENV': JSON.stringify('staging')
}),
```

Say you start afresh: run `npm start`, open a new Private/Incognito window. Go to `localhost:3998` and you'll notice that you **can't log in** via Panoptes. (Cause: the redirect URL is localhost:3998/classroom. Cause of that cause: the `env` thinks it's in staging, and staging's redirect has that extra clause added to it.)

So: 👎 

**Option 2:** we change line 48 to `'process.env.NODE_ENV': JSON.stringify('development')`

We start afresh again: run `npm start`, new Private/Incognito window. Voila, I can go to `localhost:3998` and login again.

Uncertainty: I'm not sure if this hardcoded env=development will have side effects when _we DO want to set env to staging._  But it shouldn't right? The only time we'd want to set env=staging is during `npm run deploy-staging`, and that uses `webpack.production.config.js` ? I'd appreciate a sanity check on my assumption.

So: 👍 ❓ 

**Option 3:** we change line 48 to `'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),`, same as `webpack.production.config.js` , and maybe plop a `NODE_ENV=development` to the `start` script.

So: 👍 ❓

srallen:
I think option 3 is the way to go, so that the environment is set from one place: the npm scripts. Thanks for catching this!

srallen:
@shaunanoordin I've updated to use the node environment variable set the start script. The webpack production configuration already was doing this on line 44!

shaunanoordin:
Looking good! 👍 I've also pushed a small fix for WildCam Darien's Assignment Creation, and I'll document that in my full PR review in a few mins. (EDIT: Ah, wait, I need to re-update that update a wee bit to confirm that it doesn't adversely affect I2A Assignment creation.)

PR #110, opened 2017-12-07
A few minor I2A tweaks

srallen:
This is a collection of minor tweaks:

- Updates the percentage stat for assignment completion to use rounded numbers
- Updates the production I2A program mock to correctly reflect what it is on the db
- Consistently uses the formField prop types and prop name with the `ClassroomEditor`
- Uses null for the `ClassroomEditor` form's `onSubmit` if the status is creating. I'm not sure if this totally works yet because production is having random 500s right now, which only happens after a request times out after a few seconds. I want to disable the submit button if there's a pending POST and the Grommet Button is only disabled if the onClick handler is null. I'm not sure if this is true for the handler set for the form on the Form component. We might need to move the event handling to the Button itself?


shaunanoordin:
Yup, that bug is a separate item and will need to be tackled separately. This PR LGTM! 👍

PR #125, opened 2018-06-05
WildCam Lab Classrooms - 2018 Rebuild - part 4

shaunanoordin:
## PR Overview
This PR follows #124. In this PR, we're mostly concerned with fixing bugs that prevent the Teachers from fully enjoying the Create/Edit WildCam Assignments functionality.

Updates in place:
- You can actually edit WildCam Assignments. There was a bug previously that prevented data from being saved on an update. (Create Assignment continues to work fine.)

Planned updates:
- The Assignment form should remember any data the user type into the form before they transition to the Map (to select Subjects). At this moment, if you type in an Assignment name, select subjects, then return, the Assignment name just goshdarn disappears/resets to the original.

### Important: Changes Affecting Astro Classrooms
@srallen please be aware that I've updated the `/lib/edu-api.js`'s `put()` function, since the pre-built request body structure wasn't flexible enough to support WildCam Classroom's PUT requests (i.e. I needed to edit the Relationship field as well as the Attributes field).

The changes are as follows:
- `src/lib/edu-api.js`: `put()` function no longer _implicitly_ constructs the request body structure.
- `src/ducks/classrooms.js`: `put()` call now _explicitly_ constructs the request body structure. Affects I2A's Edit Classroom functionality.
- `src/ducks/programs.js`. ditto. Affects... er, nothing, at the moment.

#### Side note
Sarah, since you mentioned you were making updates to I2A classrooms for summer, here are some useful notes, so you have fewer worries about your updates crossing with WildCam's functionality.
- Almost all the WildCam code has been moved to `src/modules/` and `src/programs/`.
  - they don't use any I2A Classroom/Assignment-specific code; so `src/ducks/classrooms.js` and `src/containers/AssignmentFormContainer.jsx`, for example, are safe to update for I2A.
- Only a few links to the shared code remains:
  - WildCam Classrooms still relies on the shared Join page.
    - NOTE: the Join page doesn't seem to auto-redirect students to their respective classrooms/assignments, although the join is still registered; this redirection issue seems to be present even at v1.0.0, but I haven't tested/debugged to be sure.
  - Common components like SuperDownloadButton and Status404 are still in shared use.
  - The "index" pages (src/ducks/reducer.js, ProgramHomeContainer) still have specific references to the Darien code.

### Status
WIP for WildCam Darien stuff.

The changes `/lib/edu-api.js` and its dependents are ready for review, @srallen .

### DIFF
--- a/src/ducks/classrooms.js
+++ b/src/ducks/classrooms.js
@@ -171,7 +171,7 @@ Effect('createClassroom', (data) => {
 // Hmm.... Effects can only take one argument?
 Effect('updateClassroom', (data) => {
   Actions.classrooms.setStatus(CLASSROOMS_STATUS.UPDATING);
-  return put(`/teachers/classrooms/${data.id}`, data.payload)
+  return put(`/teachers/classrooms/${data.id}`, { data: { attributes: data.payload } } )
     .then((response) => {
       if (!response) { throw 'ERROR (ducks/classrooms/updateClassroom): No response'; }
       if (response.ok) {
--- a/src/ducks/programs.js
+++ b/src/ducks/programs.js
@@ -274,7 +274,7 @@ Effect('createProgram', (data) => {
 
 Effect('updateProgram', (data) => {
   Actions.programs.setStatus(PROGRAMS_STATUS.UPDATING);
-  return put(`/programs/${data.id}`, data.payload)
+  return put(`/programs/${data.id}`, { data: { attributes: data.payload } })
     .then((response) => {
       if (!response) { throw 'ERROR (ducks/programs/updateProgram): No response'; }
       if (response.ok) {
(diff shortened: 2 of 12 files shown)


shaunanoordin:
## PR Update
This PR has been updated with the following major changes:
- The Edit Assignment feature now works as intended. (There were a bunch of bugs that got fixed, such as the edited info now showing properly ont he form.)
- UX/convenience improvements: when transitioning to/from the Assignment form and the WildCam Map (to select subjects), a lot of the data is now "remembered" so users don't have to re-type the information - the Assignment name, and any existing map filters, for example.

### Status
Merging and deploying.

Please note that after this, I'll take a short break from WildCam Lab development. A "part 5" should be coming later with a focus on: 1. information pages for teachers (this is minor), 2. a students page (plus hooking up Assignments to PFE, maybe, probably), 3. a replacement for Carto (boy oh boy).

PR #127, opened 2018-06-26
Run npm audit fix. Remove tests that don't apply anymore

srallen:
This is a first pass at just general updates. I updated the allowed node and npm versions, so I could run npm 6 (node 10 will probably be LTS this fall). I ran `npm audit fix` to get non-breaking updates to the packages. I removed old tests that fail and don't apply anymore that weren't updated as we were rapidly developing last fall. 

The rest of the updates that can be done will have to be done manually since they are major version updates. Most of them are related to webpack and babel.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,12 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@types/node": {
-      "version": "8.0.51",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-8.0.51.tgz",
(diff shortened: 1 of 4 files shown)


srallen:
Note: I'll be branching off this one to do the manual package updates and then for later for debugging issues until this gets merged.

shaunanoordin:
Since this PR will have sub-branches for further updates, shall we add a WIP label so I don't accidentally approve/merge the PR prematurely?

srallen:
@shaunanoordin Actually you can go ahead and merge. I'm still in the middle of working on some of these manual updates and testing. Thanks

PR #128, opened 2018-06-26
Manual package updates

srallen:
Fixes most of the remaining packages with warnings from `npm audit.` Updating webpack has a cascade of updating almost every package for webpack to get peer dependencies up-to-date and compatible.

One update to the webpack configs is to add [the `mode` configuration](https://webpack.js.org/concepts/mode/) which adds sensible built-in defaults depending on the environment.

App appears to compile and run in development and also is building, tested by running `npm run build`


webpack.production.config.js
```diff
@@ -31,18 +31,10 @@ module.exports = {
       filename: '[name]-[hash].min.css',
       allChunks: true,
     }),
-    new webpack.optimize.UglifyJsPlugin({
```
shaunanoordin: Out of curiosity, why are we removing Uglify? I see that when running `npm run build`, the output JS is still minified, but I haven't figured out _where_ this is happening.

srallen: The webpack version 4 production mode includes these plugins with their defaults:
```
Provides process.env.NODE_ENV with value production. Enables FlagDependencyUsagePlugin, FlagIncludedChunksPlugin, ModuleConcatenationPlugin, NoEmitOnErrorsPlugin, OccurrenceOrderPlugin, SideEffectsFlagPlugin and UglifyJsPlugin.
```
UglifyJsPlugin has a default to ignore IE8 support.

Issue #89, opened 2017-11-13
Join classroom bug


srallen:
Sometimes the request to join a classroom happens twice.

PR #129, opened 2018-06-28
Manage our own oauth redirects. Add a initialization loading page.

### LINKED ISSUE #121: Issues with joining classroom after signing in.
When clicking a join classroom link, I'm prompted to sign in to my Zooniverse account - but after signing in I'm re-directed to the classrooms.zooniverse.org homepage without any indication of whether or not I successfully joined. Clicking the link again opens up a new tab where I'm not signed in, leading to the same issue all over again. I'm not sure if this is the intended behavior (i.e. not maintaining signed-in status across tabs). 

The current solution is to sign in, let it redirect to the classrooms.zooniverse.org page, and then copy/paste the link into the address bar of that specific tab. Would prefer to have feedback after the sign-in completes, and/or automatically re-try to join the classroom after signing in.

srallen:
Fixes #121. This is built on zooniverse/panoptes-javascript-client#97 (up to you if you want to install from github or use `npm link` to test).

I refactored the oauth initialization to include `customRedirects: true` because this app is managing the redirects on its own. I updated the redirect code in the `AuthContainer` to store the redirect in localStorage regardless of what page you're on. I also moved the initial login check to `Main` and added a general status loading page before we know if there's a user or not which just seems like good general practice.

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { Actions } from 'jumpstate';
 import { Route, Switch, Redirect } from 'react-router-dom';
 import App from 'grommet/components/App';
 import Box from 'grommet/components/Box';
@@ -11,6 +12,7 @@ import AuthContainer from '../containers/layout/AuthContainer';
 import AppNotification from '../containers/layout/AppNotification';
 import ProgramHomeContainer from '../containers/common/ProgramHomeContainer';
 import JoinPageContainer from '../containers/common/JoinPageContainer';
+import GenericStatusPage from './common/GenericStatusPage';
 
 import AppHeader from './layout/AppHeader';
 import {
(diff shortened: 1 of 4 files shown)


shaunanoordin:
Testing note updates: derp, I forgot how _general_ the "Redirect After Sign In" feature for Edu API FE is.  The correct statement for the feature is this:

_"If a non-logged-in user visits any Page X (e.g. http://localhost:3998/#/wildcam-darien-lab/map/) in Edu API FE, and clicks Sign In, the user should be automatically redirected back to Page X upon successfully returning from the Panoptes Sign In page"_

The redirect is _not_ limited to just the Join Page.

shaunanoordin:
WHOOPS! I forgot I have a dinner appointment in a few minutes, so I'll do this quick and continue when I'm back:

- the latest update to PJC properly fixes the redirect issue - if a non-logged-in user visits Page X, then Signs In, they are returned back to Page X. WOOT! 👍 
- However, there is something weird with Test case 2 and Test case 3.
- With Test case 2, the user is redirected back to the Join page, but the "Join failed" status message is displayed instead.
  - BUT!!! The join request is actually successful (i.e. the student is listed in the classroom, as verified by the teacher owning the classroom), meaning... the Join logic was probably executed twice?
- With Test case 3, something similar happens... the user is redirected back to the Join page, the "Join success" message is displayed... but a "Something Went Wrong" toast _also_ appears.
  - The join request is actually successful (again, as above) so I'm not sure where the 'perceived errors' are coming from yet.

I'll get back to this soon!

srallen:
The issue in test case 2 and maybe 3 is logged in #89. Debugging the double join request was impossible to debug with this redirect bug, so it can be worked on after this merging.

PR #155, opened 2019-06-11
Updates ZRC. Runs npm audit fix

srallen:
Manually updates ZRC and runs npm audit fix. The few tests we have pass. I'm not sure if any manual testing should happen. In theory patch updates shouldn't break anything.


shaunanoordin:
## Quick Review

- This PR looks good, but given the shenanigans I experienced deploying zooniverse/anti-slavery-manuscripts#289 I'm going to wait until tomorrow until I've had a chance to more thoroughly test this.
- I ran `npm audit` and saw that there are still 38 vulnerabilities (23 low, 12 moderate, 3 high) that haven't been patched, so I'll address those tomorrow as well.

Issue #148, opened 2019-03-20
Data Export Problem for Classroom.Zooniverse.org Galaxy Zoo 101


trouille:
@srallen I just got an email from Rose Finn using classroom.zooniverse in her Astro101 course. She used it last year as well. 

She's run into a data export error. 

Here's her info:
my username is Rose_Finn (rfinn@siena.edu).  My class for this spring is
https://classroom.zooniverse.org/#/astro-101-with-galaxy-zoo/students/classrooms/2673/join?token=7b9a17456a3f1962

and this is the project page for galaxy zoo
https://www.zooniverse.org/projects/zooniverse/galaxy-zoo-in-astronomy-101/classify?group=1749049&_ga=2.195639468.1877613624.1553048722-149825245.1553048722

If I go to https://www.zooniverse.org/lab/2308/data-exports - is there a workflow I can choose to export data and maybe manually pull out her classrooms data? Or other way that I can try to help her (without imposing on your time this week)?

trouille:
Here's the error she gets: 
![Screen Shot 2019-03-20 at 11 31 10 AM](https://user-images.githubusercontent.com/4975655/54704258-62bebe00-4b08-11e9-8eb4-a0eedaa20e78.png)

srallen:
These are caesar exports, so this indicates and issue with those. We can follow up in those logs looking for reductions exports for that user group id and those project's workflows that failed.

srallen:
Fwiw, we can't really use Panoptes classification exports without doing manually the aggregation caesar does for those to make them usable.

trouille:
Thanks! (And good reminder that caesar manipulates the data to make it more useable - so not that helpful to get the panoptes raw classification export)

I can ask Amy to check on caesar -- but to be sure I know what to ask for (also so I know for future cases) - are these the right numbers to share with Amy?: 
workflow = 2673
group ID = 1749049
?

trouille:
(I'll share the full URLs w/ her, but it's for my knowledge as well).

srallen:
The workflow ids are:
Galaxy Zoo 101: 5521
Hubble's Law: 5522

And user group id is correct, 1749049

(id 2673 is the classroom id which is only used by edu api, not by panoptes or caesar)

trouille:
@srallen See https://github.com/zooniverse/caesar/issues/721

[...]

PR #158, opened 2019-06-27
Fix exports

srallen:
Caesar's API for requesting data exports has changed which is the likely source of the issues in #148. This is a partial fix to the issue. I've updated the requests to use `subject_reductions` in the request payload as well as changed `workflow_id` to `exportable_id` in the store. 

I'm unsure if anymore client side changes need to be made, but there are some possible server side bugs now being exposed now that the request is properly being made. See https://github.com/zooniverse/caesar/issues/859

### DIFF
--- a/src/components/astro/AstroClassroomsTable.jsx
+++ b/src/components/astro/AstroClassroomsTable.jsx
@@ -129,7 +129,7 @@ function AstroClassroomsTable(props) {
                           type="button"
                           className="manager-table__button--as-link"
                           plain={true}
-                          onClick={props.showExportModal.bind(null, assignment, classroom)}
+                          onClick={calculatedCompleteness !== 0 ? props.showExportModal.bind(null, assignment, classroom) : null}
                         >
                           Export Data{' '}
                           <i className="fa fa-arrow-down" aria-hidden="true" />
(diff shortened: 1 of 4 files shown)


src/components/astro/AstroClassroomsTable.jsx
```diff
@@ -129,10 +129,7 @@ function AstroClassroomsTable(props) {
                           type="button"
                           className="manager-table__button--as-link"
                           plain={true}
-                          onClick={calculatedCompleteness !== 0 ?
-                            props.showExportModal.bind(null, assignment, classroom) :
-                            null
-                          }
+                          onClick={props.showExportModal.bind(null, assignment, classroom)}
```
srallen: This change should be reverted before merging. It is here for making debugging easier.

src/ducks/caesar-exports.js
```diff
@@ -182,14 +182,16 @@ Effect('createCaesarExport', (data) => {
...
+      requested_data: 'subject_reductions',
       subgroup: data.classroom.zooniverseGroupId
     })
     .then((response) => {
       if (!response) { throw 'ERROR (ducks/caesarExports/getCaesarExport): No response'; }
       if (response.ok && response.body) {
+        console.log('response.ok', response.body)
         const responseData = response.body;
         if (responseData.status === 'pending') {
+          console.log('pending status')
```
srallen: Logs will be removed prior to merge. Here for debugging

srallen:
Tested with staging caesar:

- An export can be successfully initially requested with a pending response
- Second poll is successful and stores the completed export state with options to download CSV directly or export to google drive
- CSV download successful and can open in Numbers
- Sending to Drive successful, link to open in google sheets successfully generated and worked.

PR #174, opened 2019-11-19
[WIP] Fix `deploy-staging`: staging app should connect to staging API

shaunanoordin:
## PR Overview

This PR (attempts to) fix `npm run deploy-staging`, because at the moment it's building and deploying the app to `http://classroom.preview.zooniverse.org/` with _production_ specs instead of _staging_ specs. This results in unexpected behaviour, where users login with their proper Zooniverse accounts instead of their staging accounts.

Tagging @srallen and @camallen for an FYI.

### Dev Notes

- Main problem was that `npm run deploy-staging` was using the same `build` command as `deploy-production`, and that build command is hardcoded to run production specs.
- So far I've managed to split `deploy-staging` so it has its own `build-staging` command, and corresponding `webpack.staging.config`.
- Current issue I'm troubleshooting is that the webpack config file keeps overriding the NODE_ENV variable based on its `mode`, which is either 'production' (which is the default, and takes place if 'mode' isn't specified), 'development', or 'none'... and using 'none' actually sets NODE_ENV=none, which breaks the Panoptes JavaScript Client.
- Sarah pointed out that for our other projects, the solution for specifying our own ENV variables is to use `DefinePlugin`, so that's what I'm going to try next.

### Status

DEBUG MODE! WIP! DO NOT MERGE YET

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,6 +11,13 @@
       ],
       "plugins": ["transform-object-rest-spread"]
     },
+    "staging": {
+      "presets": [
+        ["es2015", { "modules": false }],
+        "react"
+      ],
+      "plugins": ["transform-object-rest-spread"]
+    },
     "test": {
       "presets": ["es2015", "react"],
       "plugins": ["transform-object-rest-spread"]
--- a/package.json
+++ b/package.json
@@ -11,8 +11,9 @@
     "test": "NODE_ENV=development BABEL_ENV=test mocha mocha-setup.js $(find src -name *.spec.jsx) --compilers js:babel-core/register --require mock-local-storage || true",
     "eslint": "eslint .",
     "build": "export BABEL_ENV=production; check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
-    "deploy-staging": "export NODE_ENV=staging; npm run build && publisssh dist zooniverse-static/preview.zooniverse.org/classroom",
-    "deploy-production": "export NODE_ENV=production; npm run build && publisssh dist zooniverse-static/classroom.zooniverse.org"
+    "deploy-production": "export NODE_ENV=production; npm run build && publisssh dist zooniverse-static/classroom.zooniverse.org",
+    "build-staging": "export NODE_ENV=staging; export BABEL_ENV=staging; check-engines && check-dependencies && webpack --config webpack.staging.config.js && echo NODE_ENV is $NODE_ENV",
+    "deploy-staging": "export NODE_ENV=staging; npm run build-staging && publisssh dist zooniverse-static/preview.zooniverse.org/classroom"
   },
   "engines": {
     "node": ">=8.1.2",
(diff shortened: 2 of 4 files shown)


eatyourgreens:
Pandora also has this problem, even if I explicitly `export NODE_ENV=staging;` before a build. 😞
https://github.com/zooniverse/pandora/pull/170

I'm using the [webpack `EnvironmentPlugin`](https://webpack.js.org/plugins/environment-plugin/), which should do the same job as the `DefinePlugin`.

eatyourgreens:
I found the answer here: webpack’s build mode overrides any value set with `DefinePlugin`. https://stackoverflow.com/questions/51163005/is-using-mode-and-webpack-environmentplugin-redundant

In the monorepo, we use a separate `PANOPTES_ENV` to specify the API environment, allowing for webpack production builds (minified code etc.) that use the staging API and auth. 

I’m kind of interested to know how PFE staging builds work now.

eatyourgreens:
I looked into this a bit more this morning. I still can't see how PFE manages to build a production webpack build that uses the staging API.

Here's the output of `console.log(argv)` for a PFE staging build:
```js
{
  _: [],
  cache: null,
  bail: null,
  profile: true,
  color: true,
  colors: true,
  config: './webpack.build.js',
  progress: true,
  'info-verbosity': 'info',
  infoVerbosity: 'info',
  '$0': '/Users/odonnell/zooniverse/Panoptes-Front-End/node_modules/.bin/webpack'
}
```

Here's `argv` for a Pandora staging build:
```js
{
  _: [],
  cache: null,
  bail: null,
  profile: null,
  color: { level: 2, hasBasic: true, has256: true, has16m: false },
  colors: { level: 2, hasBasic: true, has256: true, has16m: false },
  config: 'webpack.production.config.js',
  p: true,
  'info-verbosity': 'info',
  infoVerbosity: 'info',
  '$0': '/Users/odonnell/zooniverse/pandora/node_modules/.bin/webpack',
  'optimize-minimize': true,
  define: [ 'process.env.NODE_ENV="production"' ],
  mode: 'production'
}
```

The Pandora build explicitly defines `NODE_ENV=production`. Both builds used webpack 4.44. I can't see what's different in the webpack config.

eatyourgreens:
After some more digging around, I think the solution to this problem is to remove `-p` from the build and explicitly set build options in the webpack config file. See https://github.com/zooniverse/pandora/pull/170#issuecomment-735777225

shaunanoordin:
Thanks Jim! This PR's a bit low on the "clean up and resurrect" priority list, but discovering the `-p` flag is a definite win.

Side note: I'm comparing Webpack v4's (which EduAPIFE and PFE-lab use) CLI flags (https://v4.webpack.js.org/api/cli/) with the current v5's (https://webpack.js.org/api/cli/) documentation and I'm noticing that the `-p` flag's been removed, or at least no longer documented.

srallen:
Closing because deploy goes to Azure via GA now.

PR #179, opened 2020-01-24
add dev docker setup

camallen:
quick and easy development via docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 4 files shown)


camallen:
Looks like the fix is to install the dep manually https://github.com/imagemin/cwebp-bin/issues/31#issuecomment-588383966

most likely related to the underlying image changing a debian release and the package is upgraded / missing for that release. Might be worth investigating if we can use a later node /debian image vs `0.10-wheezy` https://hub.docker.com/_/node/?tab=tags&page=1&name=0.10-

If we can't then the first solution is fine, we just install the underlying dep for that npm package.

srallen:
I can attempt a node version bump. I don't know of any reason why that shouldn't work.

srallen:
This is working now with the node update. Thanks!

PR #192, opened 2020-05-06
Reverse spiral and elliptical for export

srallen:
We had a report that the header labels in the CSV export for the relaunched Galaxy Zoo activity were reversed. This was confirmed by comparing the tasks from the original project to the new project, that elliptical is now the first choice in the question task list.

Caesar has no knowledge of the human labels for question task answers, so when you get a reducer export, the labels are in array order of the task answer list, like `data0`, `data1`, etc. I take those labels and give them human labels again, `spiral`, `elliptical`, etc. This relabeling is hardcoded and assumed the order of the answers in the task remained the same. This PR fixes the ordering for the new project.

Long term, we should consider a few of options:
1. Zooniverse classrooms requests the workflow from Panoptes to get the human labels.
2. Caesar looks up the Panoptes workflow human labels and stores the labels along side the reduced data or at time of the reduction export
3. The workflow labels could get submitted as part of the classification so Caesar can use them in storing the extract and use it in the reduction as needed and when exports are requested.

Option 1 only benefits GZ project on Zooniverse classrooms (we don't know yet if other classroom projects will need this) and is front-end work. Option 2 or 3 is a broader impact and will take more (back-end) dev time, but possibly of interest to many projects to enhance the exports from Caesar.

### DIFF
--- a/src/containers/astro/AstroClassroomsTableContainer.jsx
+++ b/src/containers/astro/AstroClassroomsTableContainer.jsx
@@ -92,14 +92,14 @@ class AstroClassroomsTableContainer extends React.Component {
   }
 
   transformGalaxyDataCsv(csvData) {
-    let csvRows = 'Galaxy ID,Total # of classifications,Spiral,Elliptical,Merger,Artifact,SSDS ID,Image,GZ Original Spiral,GZ Original Elliptical,GZ Original Merger,GZ Original Artifact,Computer Classification\n';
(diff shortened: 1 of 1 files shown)


lcjohnso:
Following @srallen's suggestion that a general solution is preferred, I also prefer either Option 2 or 3.  Option 3 sounds like a heavy-weight solution so not favorable - my understanding is these labels are only attached during data export phase, would create larger payloads to send to Caesar (and the extra content is typically unnecessary), and storing along with classification sounds inefficient.

Option 2 sounds attractive for multiple reasons:
- Only fetch the current up-to-date label info when necessary, saving on data storage.
- Building the infrastructure to query Panoptes for relevant labels would also be useful for UI use when rule-writing.  e.g., currently must define rules based on answer index for question task reductions -- would be nice to show what label that matches to, even if only for confirmation purposes.
- I would suggest a "at time of export" implementation -- matches my understanding of how Panoptes data exports also function.

PR #199, opened 2020-06-30
Update ZRC, add dev dependency babel register, run npm audit fix

mcbouslog:
Update Zooniverse-React-Components to v0.9.1, which includes security vulnerability patches, but no changes.

On other repos I've been using this as an excuse to also run an `npm audit fix`, let me know if would also like me to open that PR, or I'd be happy to review that PR, and get both in for a deploy.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2036,6 +2036,19 @@
         "@babel/plugin-transform-react-jsx-source": "^7.7.4"
       }
     },
+    "@babel/register": {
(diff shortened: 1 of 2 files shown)


mcbouslog:
Oh prob for another day, but had to `npm install @babel/register --save-dev` for `npm test` to run. Kudos for any tests, don't look at NfN Field Book! ;-)

PR #215, opened 2020-12-22
Update react, enzyme, and sinon

srallen:
First step in updates: React to version 16 which requires enzyme updates to be able to run tests. I also updated sinon to  address some of the test failures. The dependecies sinon-test and react-test-renderer were able to be removed due to the updates too.


shaunanoordin (approved):
## PR Review

- This PR is a major dependency update. 
- Testing looks good - I cleared my modules, reinstalled, and tested all the baseline functionality (loading pages, navigation, sign in/out) of the website on macOS10+Chrome on localhost.

I don't see any issues with the jump from React 15 to 16, so this LGTM 👍

PR #223, opened 2021-03-22
WildCam Lab docs 2021

shaunanoordin:
## PR Overiew

This PR adds additional documentation and guides for setting up and updating WildCam Labs.

- New folder: 'docs/wildcam/misc-tools'
- New script: `filter-csv.py`. This script reads a CSV file, finds rows that meet a specific criteria, and copies those rows to an output CSV. Useful for, e.g., trimming down mega huge Zooniverse Classification exports, so we only have Classifications from specific Workflows.
- TODO

### Status

WIP


camallen (commented):
I feel like these docs should be on the https://github.com/zooniverse/classroom-maps-api/ repo, especially as that includes a python run time via docker.

docs/wildcam/misc-tools/filter-csv.py
```diff
@@ -0,0 +1,37 @@
...
+
+with open(inputFilename, mode='r', newline='') as inputFile, \
+     open(outputFilename, mode='w+', newline='') as outputFile:
+  
+  inputReader = csv.DictReader(inputFile)
+  outputWriter = csv.DictWriter(outputFile, fieldnames=inputReader.fieldnames)
+  outputWriter.writeheader()
+  
+  for inRow in inputReader:
+    if inRow[columnName] == columnValue:
```
camallen, line 35: you might want to compare the runtimes of this python script vs cmd line tools like `awk`, https://github.com/camallen/cmd-line-data-goodies/blob/master/examples/12_awk.md

```
# use a regex to search for values in column 3
# and print the row num NR and the row $0
awk -F, ' $3 ~ /columnValue/ { print NR, $0 } ' $inputFile

# use a direct string match for values in column 3
awk -F, ' $3 == "columnValue" { print NR, $0 } ' $inputFile
```

shaunanoordin:
Thanks Cam!

- Good point, the current doc might actually make more sense on `classroom-map-api`. I'm still gathering my thoughts on this though since I need to expand this README with an additional section called "how to copy Classroom's wildcam-lab modules to make a new Program."... I might end up with two READMEs (one for Classrooms and one for the Classrooms Maps API).
- `awk`?? Oh man, I need to spend some time grokking these command line tools. Python is my default for file-related scripting at the moment. (Though this is admittedly much better than me relying on Node.js for the equivalent.)

I'm going ahead and merging these doc updates first so I have a save point while I work on the "spinning off a new Program" bit.

PR #226, opened 2021-03-29
Rename image extensions to use lowercase

srallen:
OS X is case insensitive in ways that every other flavor of Linux is not. This should fix the failing Github Actions for deployment.


srallen:
@shaunanoordin be sure that the import of any static asset matches the case in the file system. Thanks!

shaunanoordin (approved):
## PR Review

This PR changes the filenames of a few Darien .JPG (now .jpg) files. LGTM 👍 

Sarah: noted on ensuring that 1. all future filenames are lowercase, and 2. import statements match the lowercase-ness. This will be pretty important for WWKenya, thanks for highlighting this.

PR #242, opened 2021-06-09
Jun 2021 hackday: add optional Tooltips for layers

shaunanoordin:
## PR Overview

Affects: WildCam Lab's Map Explorer (i.e. Gorongosa, Darien, and Kenya)

Context: ever since Wildwatch Kenya Lab, we realised that the "extra data layers" on the map (the ones that represent biomes/habitats, park boundaries, etc) actually have a lot of important information that can be surfaced.

This PR adds an optional Tooltip feature. By mousing over certain Features on those Extra Data Layers, a Tooltip will appear displaying all the information we have on that map feature.

Screen Shot 2021-06-09 at 15 46 39

- A basic "bindTooltip" call has been added. This creates a default tooltip.
- NOTE: for an extra data layer to have a tooltip, it needs to have its **layer.style.interactive** value set to **true.**
- Required secondary functionality: now every time a layer is added/removed (via the built-in layer controls), the Data Layer (aka Camera Layer) is ALWAYS brought to front.
  - Otherwise, we wouldn't be able to click on the cameras. Any layer with layer.style.interactive=true will block interactions to layers beneath it.

### Status

Technically ready, but functionally a WIP. This is practically complete but I'm thinking of adding one or two more flourishes, if I can spare the dev time:
- the tooltip should appear close to the mouse, instead of centred on the feature, which can be unreliable at different levels of zoom.
- when the tooltip appears on a Feature, that Feature should have some sort of highlight, e.g. bold borders.

If I can't spare the dev time, I'm going to have to get this merged as-is by Friday so I can deploy the latest Kenya map updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "debounce": "~1.0.2",
         "grommet": "~1.13.0",
         "jumpstate": "~2.2.2",
-        "leaflet": "1.2.0",
+        "leaflet": "1.7.1",
         "panoptes-client": "~3.1.0",
         "papaparse": "~5.2.0",
         "prop-types": "~15.7.2",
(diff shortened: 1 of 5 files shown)


shaunanoordin:
## Update

I'm going to go ahead and merge this first before this branch goes stale; the planned enhancements will be taken care of in a separate PR.

TODO:
- Improve UI for tooltips. i.e. "centre tooltip on mouse", not "centre on Feature"
- Optional stylings for tooltip-ed Feature, i.e. bold borders
- Add new maps for Kenya (this include correcting the incorrectly named "Kenya" layer)

PR #249, opened 2021-10-05
Add `build-staging`

shaunanoordin:
## PR Overview

This PR adds the `build-staging` script, so our [staging website](https://classroom.preview.zooniverse.org/) uses Staging instead of Production data.

- Currently, https://classroom.preview.zooniverse.org/ connects to Production because there's only one `build` script, and that's for prod.
- Note that the staging website is auto-deployed via Github Actions

### Status

Code ready to review, but mind you, I can't seem to get `npm run build-staging` (nor `npm run build`) on my machine for self testing, due to a bunch of configuration shenanigans. Sigh, don't ask, I keep getting Python errors even though this shouldn't be using any Python. (`npm run start` works fine though, so I know the website code is working OK, it's my local machine config that's borked.)

@camallen Can you please take a look at this? Especially the line...

```
"build-staging": "export BABEL_ENV=staging; check-engines && check-dependencies && webpack --config webpack.config.js -p"
```

...I'm not sure if that should be webpack.production.config.js , because as I said, I can't compile locally. 😢

PR #250, opened 2021-10-05
Fix build-staging for Webpack 5

shaunanoordin:
## PR Overview

PR #249 failed in its intent - `npm run build-staging` was failing. PR #248 should have been merged first, since there were essential Webpack 5 upgrades. This PR attempts to fix 249.


shaunanoordin:
It works! https://classroom.preview.zooniverse.org/#/

The error I made was assuming that env `staging` == env `development`, when in fact there are situations (e.g. on the Babel config) where 'development' is defined, but not 'staging'. Changing all references to env `development` did the trick.

PR #252, opened 2021-11-05
Fix az deploys

camallen:
this pr specifies a working az cli version to avoid this issue https://github.com/Azure/azure-cli/issues/20154

Linked to https://github.com/zooniverse/Panoptes-Front-End/pull/6044

PR #254, opened 2021-11-23
Remove azcliversion from github actions

mcbouslog:
Per azure-cli/issues/20154, `azcliversion` should be omitted so defaults to latest, or set to latest. For consistency with other Zooniverse repos this PR removes `azcliversion: 2.29.2`.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -28,7 +28,6 @@ jobs:
     - name: Upload to blob storage
       uses: azure/CLI@v1
       with:
-        azcliversion: 2.29.2
         inlineScript: |
             az storage blob upload-batch \
               --account-name zooniversestatic \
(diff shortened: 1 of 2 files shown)


mcbouslog:
@shaunanoordin - see https://zooniverse.slack.com/archives/C01M8PJGLR5/p1637604211090100 for additional context. I don't think there's anyway for you to test this, I think we just have to merge and monitor?

shaunanoordin (approved):
## PR Review

This PR fixes some deployment issues, and follows the "temporary workaround fix(?)" of #252 (which added azcliversion).

LGTM, I'm happy to get this merged so we can monitor that the deployment is fixed. I'll keep an eye on that.

Side note: prior to 252, we had `creds: ${{ secrets.AZURE_STATIC_SITES }}` in place of `azcliversion`, so if the deploy fails for whatever reason, I'll check with Cam if we need to reinsert those `creds`.

Let's gooooo

shaunanoordin:
Staging deploy works fine! Wooo

PR #261, opened 2022-02-10
Build with node 16.

eatyourgreens:
#259 bumped the npm version to 7. This upgrades Node to match.

Builds are broken at the moment because Node 14 bundles with npm 6. This should fix that by building with Node 16/npm 8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16-alpine
```
eatyourgreens, line 1: Is the docker image used any more? Local development maybe?

shaunanoordin: `16-alpine` is [still listed as supported](https://hub.docker.com/_/node), so I assume it's still working? Let me run this on Docker local to see if there are any hiccups, but it should be no worries.

eatyourgreens: I've changed it to `node:16`.

PR #273, opened 2022-04-27
Use shared CI-CD workflows for builds and deploys

eatyourgreens:
- Upgrade the staging and production deploys to use shared workflows.
- Run a test build on PR branches, to check that we aren't breaking the build.


mcbouslog (approved):
compared to Pandora, LGTM

PR #309, opened 2022-04-28
Fix broken production build

eatyourgreens:
An update to `ImageMinimizerPlugin` broke the production webpack build. This updates the plugin config to match the latest version.


eatyourgreens:
Production builds are running again. 🎉

PR #318, opened 2022-05-02
Add Activities for Undergraduates program home and related

mcbouslog:
Per https://docs.google.com/presentation/d/1XaxWqMLs_9Cudh2Eg6s3TUsMIX7kQj_2z4Cy-jOvT8M/edit#slide=id.g109c30b4a63_0_132.

Adds new Activities for Undergraduates program, including:
- add mock programs data
- update routes
- add home page
- update nav


shaunanoordin:
To repeat, the notes above ☝️ are just for your information, I have no actual review/comments/recommendations for this draft PR. 😄 

For historical context:
- Both WildCam Darien Lab and Intro2Astro used to store code in `src/containers/...` and `src/components/...` and `src/ducks/...`. These were the original 2 programs of Classrooms (née Education Front End)
- However, when we started adding WildCam Gorongosa Lab, I realised that the code would get very very messy with 3 different programs all sitting and mingling in the `src/containers/...` and `src/components...` and `src/ducks/...` folders
- So, WildCam Darien Lab's code was extracted from those folders to `src/programs/darien`
- Then, all the reusable Darien classroom code was extracted from that folder to `src/modules/wildcam-classrooms` (which, you'll note, now has its own `/components`, `/containers`, and `/ducks` subfolders) (Basically, same code, just moved a few folder levels down, for organisational purposes.)
- Gorongosa (and later Kenya) would follow this pattern.

mcbouslog:
Thank you @shaunanoordin for the context and help! The Activities for Undergraduates program is indeed similar to Intro2Astro, even simpler, in that its solely links to Google Doc folders for a few specific classes.

mcbouslog:
This new program doesn't have an actual Zooniverse project associated with it. I think I see that in **education-api** a `slug` property is required on a `program`. Does it need to be a real project slug, or can I define it and never use it? Do I need to create a related project (even if it's never used)? Not a problem to do, just checking.

This new program is really just static information (names, descriptions, etc., and links to Google Drive folders). Should I not make a program for it in **education-api** and hard code where necessary here in `classroom`? I don't like the sound of that, but I'm not sure creating a program in the db is really the best approach either.

@shaunanoordin your thoughts appreciated.

mcbouslog:
Still open to suggestion, but I'll get help creating a program as typically done.

To test locally, I suggest changing [programs line 300](https://github.com/zooniverse/classroom/blob/zoo-undergrads/src/ducks/programs.js#L300) to `const sortedPrograms = sortPrograms(programsArray);`.

shaunanoordin (approved):
## PR Review

The PR adds a new education Program: "Activities for Undergraduates"

- Program details:
  - AfU is a standard content page. 👍
  - URL will be is `https://classroom.zooniverse.org/#/zooniverse-based-activities-for-undergraduates`
  - AfU is accessible from the home page ⚠️ but only from _staging,_ for now.
- External dependencies: new `program` resources has been added to Education API staging, ⚠️ but needs to be added to **production** as well

### Dev Notes

@mcbouslog I'm sorry that I didn't answer your questions earlier! OK, so the questions were...

**Q1:** should you add the Program as a hardcoded item in `ducks/programs.js`, or add it as a resource to the Education API?

A1: you should do _both._ IIRC, the hardcoded items in ducks/programs.js are the fallbacks (in case Education API is down, but you still want some basic functionality on the website). The code still relies on a lot of  data from the Education API. Sorry, it's a bit messy.

- [ ] **[major] Suggestion 1:** you should add the AfU Program to Education API **production** as well. See the Testing section for why.

**Q2:** do you need to add a `slug` to the program?

A2: Yup, you _should,_ if only for consistency. `program.slug` needs to be the _correct path/route to the program,_ and is used by (for example) the Join Classroom link:

```
joinURL = `${config.origin}/#/${props.selectedProgram.slug}/students/classrooms/${props.selectedClassroom.id}/join?token=${props.selectedClassroom.joinToken}`
```

- [ ] **[very minor] Suggestion 2:** if you haven't advertised the URL yet, you could shorten the AfU slug from `zooniverse-based-activities-for-undergraduates` to `activities-for-undergraduates ` . This is purely a subjective aesthetic thing, since the full URL will have the word Zooniverse twice : `https://classroom.zooniverse.org/#/zooniverse-based-activities-for-undergraduates`
  - Warning: if you do decide change this, you'll need to update the slug accordingly in four places: 1. ducks/programs.js , 2. ProgramHomeContainer.jsx , 3. Education API Staging, 4. Education API Production.
  - Personal note: don't worry about Suggestion 2 if changing the program.slug for an existing is a massive faff. I think I remember I ran into an issue where my account had permissions to _create_ Programs on EduAPI, but not _edit_ them, so I ended up having to ask Zach or Cam for help. 🤔 That problem might be solved now, but man, it bugged me because I just had one small change to make and it took a whole day to resolve.

### Testing

Testing AfU is pretty easy since this is a content page with no special UI/UX features/functionality, but it's important to check for both _staging_ and _production_

Staging:
- AfU is listed on the home page - https://local.zooniverse.org:3998/#/
- AfU Program page looks good - https://local.zooniverse.org:3998/#/zooniverse-based-activities-for-undergraduates
  - Text content, images, and links all check out 👍 

Production:
- ❗ AfU is **not** listed on the home page - https://local.zooniverse.org:3998/?env=production#/
- ❕ AfU Program page looks mostly good, but is missing the header image - https://local.zooniverse.org:3998/?env=production#/zooniverse-based-activities-for-undergraduates

### Status

This PR gets a thumbs up 👍 from me, though I _highly recommend_ addressing Suggestion 1 above before you merge and deploy. Let's gooo!

src/ducks/programs.js
```diff
@@ -137,7 +162,7 @@ const gorongosa = {
 
 const kenya = {
   staging: {
-    id: '5',
+    id: '6',
```
shaunanoordin, line 165: Wait, you're right, Kenya IS Program 6 on staging but Program 5 on production. Goshdangit, I think I remember that this might have been my fault - I could create a new Program _but not edit it,_ so when I found a mistake, I think I had to delete Program 5 on staging then create a new Program 6 with the correct details this time. This lack of synchronisation is going to haunt my goshdang dreams. 🤨

mcbouslog:
Thank you @shaunanoordin ! Agreed regarding the suggestions, but I think I'll merge this so it'll be on staging so I can share with Laura and Molly for any additional edits. With their approval I'll also add the program to the production API before deploying to classroom to production as well.

PR #323, opened 2022-05-13
Refactor build scripts

mcbouslog:
- refactor build scripts, primarily so staging uses `webpack.production.config.js`
- updates GitHub workflows for new build scripts

### DIFF
--- a/.github/workflows/build_branch.yml
+++ b/.github/workflows/build_branch.yml
@@ -13,4 +13,4 @@ jobs:
       commit_id: ${{ github.sha }}
       node_version: '16.x'
       output: 'dist'
-      script: 'build'
+      script: '_build-staging'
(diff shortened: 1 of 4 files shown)


mcbouslog:
Used https://github.com/zooniverse/pfe-lab/blob/master/package.json as a reference, but open to suggestions.

eatyourgreens:
`NODE_ENV` conflicts with the Webpack build mode in the staging script. That’s confusing. 
https://webpack.js.org/configuration/mode/

eatyourgreens (approved):
Let’s see how this works. We can always iterate on the settings.

mcbouslog:
Agreed that doesn't look right with the staging env and production mode. But sounds good on revisiting too, I'll merge this to see if works. Thanks @eatyourgreens !

PR #326, opened 2022-05-14
Revert "Bump react-redux from 5.0.7 to 8.0.1"

### LINKED ISSUE #325: ‘Create a classroom’ fails without a visible error
Pressing ‘Create a classroom’ causes the page to hang. 

From the dev console: the store is undefined in the classroom form. The fix might be the same as the fix for tutorials and field guides in https://github.com/zooniverse/Panoptes-Front-End/pull/6133

eatyourgreens:
Reverts zooniverse/classroom#284

Revert `react-redux` back to 5 so that ‘Create a classroom’ works again. 

Closes #325.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "react": "~16.14.0",
         "react-copy-to-clipboard": "~5.1.0",
         "react-dom": "~16.14.0",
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Reverting breaks the build. ☹️

PR #331, opened 2022-05-16
Fix Redux store context for Layer

### LINKED ISSUE #325: ‘Create a classroom’ fails without a visible error

eatyourgreens:
Wrap the connected assignment form and classroom form in `Layer`. This makes sure that the Redux store context is passed down correctly to the layer contents.

Closes #325.

### DIFF
--- a/src/components/classrooms/ClassroomEditor.jsx
+++ b/src/components/classrooms/ClassroomEditor.jsx
@@ -7,7 +7,6 @@ import Box from 'grommet/components/Box';
 import Button from 'grommet/components/Button';
 import Heading from 'grommet/components/Heading';
 import Label from 'grommet/components/Label';
-import Layer from 'grommet/components/Layer';
 import List from 'grommet/components/List';
 import ListItem from 'grommet/components/ListItem';
 import Paragraph from 'grommet/components/Paragraph';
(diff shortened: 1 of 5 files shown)


mcbouslog:
When on https://localhost:3998/#/astro-101-with-galaxy-zoo/educators/ signed in with my staging user, should I be able to click the "Create New Classroom" button or is there additional settings my user needs? At the moment, the button is disabled for me.

shaunanoordin:
I'm popping into this for a bit to debug that mysterious "disabled Create New Classroom" button. I'll report findings shortly.

shaunanoordin:
Good news, this is a simple fix 👌 

- The "Create New Classroom" button is disabled because its onClick prop is `undefined`
- The onClick prop is undefined because, in this PR, the `ClassroomManager.props.toggleFormVisibility` value is undefined.
- This, in turn, is because this line in `ClassroomsManager.jsx`'s `ClassroomManager.defaultProps` was removed: `toggleFormVisibility: Actions.classrooms.toggleFormVisibility`

Re-adding that _Actions.classrooms.toggleFormVisibility_ line to _ClassroomManager.defaultProps_ should solve the problem. 👍

mcbouslog (changes requested):
Thanks @shaunanoordin ! I confirmed your suggestion fixes my noted issue.

mcbouslog (approved):
Thanks @eatyourgreens ! Sorry I missed your commit a few minutes ago.

eatyourgreens:
Joining a classroom is broken, but that's because the token is in the URL fragment, not the query string.

eatyourgreens:
`joinToken` might be broken on master though, and not related to this PR.

PR #342, opened 2022-05-17
Define NODE_ENV for build script

mcbouslog:
Per https://github.com/zooniverse/classroom/blob/master/src/lib/config.js environment is defaulting to `staging` for `build` script.

This PR defines the node environment to `production` for the `build` script. 
This will change the staging classroom site to point to the production Panoptes database. Is that what we want or shuold we revisit #330 and #323?


mcbouslog:
I'm going to merge and deploy to fix the production https://classroom.zooniverse.org/#/.

PR #586, opened 2023-02-24
Upgrade webpack-dev-server

eatyourgreens:
- `npm update webpack-dev-middleware` to fix a crashing bug in `memfs` 3.4.1.
- allow hostnames other than `localhost`.
- open the dev site automatically in your default browser.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11120,12 +11120,12 @@
       }
     },
     "node_modules/memfs": {
-      "version": "3.4.1",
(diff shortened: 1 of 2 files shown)


shaunanoordin (approved):
## PR Review

This PR has some updates for local development.

- Fix: previously, on Node 18, you couldn't access the website via `local.zooniverse.org:3998` (or any localhost alias), and had to use `localhost:3998`
- New behaviour: once the dev server starts, the website automatically opens in the default web browser (in my case, Chrome)
  - Do note that it opens _localhost:3998_

